PR #19, opened 2022-05-24
Use std::string for Protein name

objarni:
Hey podock maintainer! Stumbled upon your podock repo, and really liked the philosophy of it!

This pull request suggests to switch to std::string instead of raw c string for the name member of Protein. This means you don't have to allocate/deallocate "by hand" anymore for that specific member of the Protein class.

(Also, added a .gitignore file since I by mistake added to many files to the git commit at first attempt.)

Hope you like it!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+# Ignore build arfifacts
+*.o
+*test
+metal
+podock
+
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -0,0 +1,6 @@
+# Ignore build arfifacts
```
objarni, line 1: Artifacts*, typo :)

primaryodors (approved):
Thank you so much for this PR and the kind words!

PR #21, opened 2022-05-29
Mark printable const

objarni:
- since Point|Vector ::printable does not modify 'this' state, they can be marked const
- squashes errors/warning for gcc/clion building at my machine


ssepeq (approved):

ssepeq:
Good eye catching that, thank you!

Issue #22, opened 2022-05-29
Automatic tests for point/vector missing


objarni:
Thanks for adding me as contributor @ssepeq !

Here's something I'd love to help out with:

While the `point_test.cpp` does a good job of verifying (by human eye inspection) the function of point, it would be even more valuable if the checks were automatic, so that refactoring is made easier by catching errors without resorting to human memory. It also serves as a form of documentation of the domain problem/solution.

Suggestions:
 - add automatic tests (checks) for point/vector
 - prefer modern C++ tooling for this, like doctest and CMake
 - unit tests are good enough, but approval tests may be even easier to maintain.

Is this something that would be of interest?

ssepeq:
You are very welcome! Thank you for helping out!

Regarding CMake, some of the less time intensive POdock functionality is used on the Primary Odors website, and my web host unfortunately doesn't provide CMake or let me install it, so we have to keep this project compatible with g++ I'm afraid. :( Sorry about that. If there's a way to detect whether CMake is installed and fall back to gcc without it, then that might work.

Unit tests would certainly be an improvement over my various _test.cpp apps. I don't know how to write unit tests, so it would be a welcome learning experience.

As far as the point checks, would there be any noticeable change in performance? Only because the Point class is used really extensively in various loops and a single dock can take half an hour or more depending on the number of poses and whether there's a path for the ligand to follow.

objarni:
Alright, I'll try something using only doctest and makefile.

There is a "trick" for low-tool regression tests which is based on this workflow:

- save the output of "report tests" like the _test.cpp apps you have in the repository. E.g saving point_test 
- to  point_test.approved.txt
- then, when running a specific makefile target, overwrite those reports
- if there are differences, git diff will show them.

This is a bit of a hack, but it would mean:

- no need for CMake
- no need for doctest library
- no need for approvals library

On the downside, it's not quite known/non standard, but it does the job. 

Of course, we can have both: a regression test/target with the above workflow, but also unit tests for better documentation and learning about unit tests.

ssepeq:
All of this sounds great!

objarni:
I'll start with a point_report makefile target, no extra dependencies, and enough to get a 'refactoring shield' up for point/vector (at least if point_test.cpp prints out enough information and tests enough functionality)

ssepeq:
[...]

PR #23, opened 2022-05-29
Fix issue #22

objarni:
This is a 'low tooling' fix for issue #22.

It adds a new makefile target 'point_report' which writes the output of point_test to `point_test.approved.txt`. This 'gold standard' for the point test is added to the repository, and serves as a 'regression shield' for any changes to point/vector classes, as any such change would show up when invoking git diff after running point_report. So the workflow for changing point.cpp/h would be:

1. Make changes to `point.cpp|h` (and possibly also `point_test.cpp` if new functionality)
2. Run `make point_report`
3. Run `git diff`
4. If there is a diff, something broke. Throw away changes from (1).
5. If there is no diff, all is good, commit changes!

Note: if new functionality is added to point/vector, we need to keep `point_test.cpp` up to date, as well as the `point_report.approved.txt` file which needs to be overwritten to show the new functionality.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ podock
 *.sdf
 *.txt
 *~
-
+cmake-*
+.idea/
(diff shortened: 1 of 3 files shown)


ssepeq:
Awesome!

PR #35, opened 2022-05-30
Pure virtual base and multiple inheritance.

ssepeq:


ssepeq:
Not going to continue forward in this direction.

Issue #25, opened 2022-05-29
Incorrect peptide structure from protein_test.


ssepeq:
`./protest TTTTTTTTTTTTTTTTTTTTTTT`

The result is a coiled strand where the atoms are too close together. This conformation would be impossible in a real protein.

![polythreonine_wrong](https://user-images.githubusercontent.com/66374250/170890578-ab87cb3f-9d4c-45f4-8c2b-c96223fba8cd.png)

There is a function call in protest.cpp to straighten the chain, but every iteration shows zero as the molecular energy, so no bonds are rotated.

```
Iteration 0
0 v. 0
0 vs. 0
Iteration 1
0 v. 0
0 vs. 0
Iteration 2
0 v. 0
0 vs. 0
Iteration 3
0 v. 0
0 vs. 0
Iteration 4
0 v. 0
0 vs. 0
Iteration 5
0 v. 0
0 vs. 0
...
```

ssepeq:
This is happening though.

![mangled tyrosine](https://user-images.githubusercontent.com/66374250/170895311-bcc97a68-c9bb-4812-8b73-447d4204e302.png)

ssepeq:
Alpha helices are not actually working.

ssepeq:
This is all kinds of wrong:
![peptide_wrongness](https://user-images.githubusercontent.com/66374250/171042149-30601f96-4ae1-40bd-b172-d88bf84c4168.png)

ssepeq:
The floating side chains are happening because of the proline.

ssepeq:
The segfault that is currently failing protest:

`clear; make; valgrind ./protest ARNDCEQGHILKMFPUSTWYV`

```
==321411== Process terminating with default action of signal 11 (SIGSEGV)
==321411==  Access not within mapped region at address 0x54
==321411==    at 0x10E3D0: Atom::get_location() (atom.cpp:403)
==321411==    by 0x12E24F: AminoAcid::rotate_backbone_abs(bb_rot_dir, float) (aminoacid.cpp:933)
==321411==    by 0x12B9B5: AminoAcid::AminoAcid(char, AminoAcid*) (aminoacid.cpp:162)
==321411==    by 0x1301CF: Protein::add_residue(int, char) (protein.cpp:48)
==321411==    by 0x1305CB: Protein::add_sequence(char const*) (protein.cpp:103)
==321411==    by 0x10BCFF: main (protest.cpp:36)
==321411==  If you believe this happened as a result of a stack
[...]

PR #50, opened 2022-06-04
Efforts towards bond angles. #5 #25

ssepeq:
#5 #25


ssepeq:
So far everything seems to work except rings. Doesn't matter what kind of ring, how many members, aromatic or not, bond cardinality, every single one fails every time.

Issue #69, opened 2022-06-18
Protein edit scripting.


ssepeq:
This is a huge feature which will warrant its own PR.

Create and document a scripting language for editing protein structures. Allow for parts to be moved into position, backbone chains broken and reattached, helices generated, and metal ions coordinated. The scripting language will be used to give ORs a metal-bound EXR2 helix and an EXR2 "shelf" like the EXR2 helix and "shelf" of TAAR proteins; these metal bound protein models will then be used with path-based docking to track the ligand from the extracellular domain to the pocket center.

The existing metal.cpp will be done away with and a new executable target will be created to act as an interpreter for the scripting language. The interpreter will be run with at least 2 parameters, an input script file and an input PDB file. It will be capable of at least the following functions:

- Load and save PDB files;
- Create helices and straightened sections from parts of the protein;
- Move and rotate parts of the protein in space;
- Line up parts of the protein relative to other parts; 
- Reconnect broken strands;
- Create and assign simple variables, do simple mathematical operations, and use variables as function parameters;
- Control structures: conditionals, loops, and goto;
- Write to stdout;
- Fuzzy-search the protein sequence for known motifs.

ssepeq:
Ultimately this scripting language will also enable homology modedling e.g. to create 3D models for the VN1R receptors that currently have not been modeled.

PR #70, opened 2022-06-18
Protein edit scripting. #69

ssepeq:
Initial work for ability to use variables as one-piece expressions.


ssepeq:
TODO:
- Document every command.
- Implement more robust error checking and reporting.
- Merge into main branch.
- Fix the connect command to make it actually connect broken strands.

Issue #67, opened 2022-06-15
Guided docking.


ssepeq:
Currently there is a "best binding" feature that pairs the most active moieties of ligands with the side chains most likely to bind strongly to them. That feature was built for issue #30.

Here's an article where the researchers docked an odorant to a receptor, identified active binding residues, and confirmed the role of these residues by creating mutant genes and measuring their sensitivity: https://pubs.acs.org/doi/full/10.1021/acs.orglett.2c01635

There are several articles like that in the literature. Using information about a receptor and a ligand, it could be useful to dock other similarly shaped ligands in analogous poses and compute the differences between them.

There will be a new optional guide pair parameter in the .config file that effectively creates a "best bind" pair by specifying:
- A receptor residue;
- (Optionally) a heavy atom of the residue;
- (Optionally) a heavy atom name or element symbol of the ligand;
- The ligand atom can be specified even if the side chain atom is not specified.

Up to three guide pairs will be allowed; further guide pairs will generate a warning. If the ligand atom does not exist in the ligand, or a specified side chain atom does not exist in the residue, the line will be ignored and a warning will be output. If less than three guide pairs are given, the code will use its existing algorithm to search for the remaining "best bind" pairs up to a total of three.

PR #84, opened 2022-07-10
"Tumble spheres" concept.

ssepeq:


ssepeq:
Now this is more like it!!

1A1 LIMN itsabouttime

ssepeq:
TODO:

- Create a switch to choose between the bestbind method and the tumble sphere method. Guided docking (see #67) will require the bestbind algorithm.
- d-Limonene still fails to dock in OR1A1, but visually it looks fine. The penalty for clashes is too high and/or the docker must allow interatomic distances slightly less than the sum of vdW radii.

ssepeq:
The new algorithm gives superior results for TAAR8/cadaverine vs. the best binding method. Tumble spheres should therefore always be the default unless best bind is necessary.

ssepeq:
The problem of some side chain atoms staying in their original locations when the chain flexes, is recurring.

ssepeq:
Bonds between atoms are being formed correctly during SDF load, but at some point some of the bonds are subsequently lost. This is happening in both side chains and in the ligand, and it's the reason for the left-behind atoms and the limonene fail; the molecule class incorrectly counts the bonded atoms as clashes.

ssepeq:
Tryptophan's CH atom is incorrectly labeled CH2, and that is causing the failure to flex it and its hydrogen.

Issue #89, opened 2022-07-22
Amides should remain coplanar.


ssepeq:
Asparagine and glutamine side chains are not always keeping a coplanar CONH2 moiety.

![bad ASN](https://user-images.githubusercontent.com/66374250/180461034-c817371a-2324-4686-8e18-4f4822e51cbc.png)

PR #90, opened 2022-07-23
Copied code from the nightmare branch.

ssepeq:
When new features added caused docking failures, preposterous results, and ligands that mysteriously danced around by unseen forces, it seemed all hope was lost.

This PR restores (most of) those new features with two difficult test docks intact.

### DIFF
--- a/src/classes/aminoacid.cpp
+++ b/src/classes/aminoacid.cpp
@@ -420,6 +420,7 @@ AminoAcid::AminoAcid(const char letter, AminoAcid* prevaa)
     	if (N) new_atoms[l++] = N;
     	if (HN) new_atoms[l++] = HN;
     	
+    	minimize_internal_clashes();
 		for (i=0; atoms[i]; i++)
 		{
 			if (atom_Greek[i] <= 0) continue;
@@ -443,6 +444,9 @@ AminoAcid::AminoAcid(const char letter, AminoAcid* prevaa)
 			}
 			atoms[i]->clear_geometry_cache();
 			atoms[i]->clear_all_moves_cache();
+			
+			float r = atoms[i]->distance_to(CA);
+			if (r > aa_defs[idx].reach) aa_defs[idx].reach = r;
 		}
     	
     	if (C) new_atoms[l++] = C;
--- a/src/classes/aminoacid.cpp
+++ b/src/classes/aminoacid.cpp
@@ -997,6 +997,14 @@ Molecule** AminoAcid::aas_to_mols(AminoAcid** aas)
 	return mols;
 }
 
+std::string AminoAcid::printable()
+{
+	stringstream s;
+	s << *this;
+	return s.str();
+}
+
+
 void AminoAcid::save_pdb(FILE* os, int atomno_offset)
 {
     int i;
--- a/src/classes/constants.h
+++ b/src/classes/constants.h
@@ -95,7 +95,6 @@
 
 
 // Debugging stuff.
-#define allow_auto_hydroxy 0
 #define allow_axial_tumble 1
 #define allow_bond_rots 1
 #define allow_drift 1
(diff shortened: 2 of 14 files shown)


src/classes/aminoacid.cpp
```diff
@@ -997,6 +1001,14 @@ Molecule** AminoAcid::aas_to_mols(AminoAcid** aas)
 	return mols;
 }
 
+std::string AminoAcid::printable()
+{
+	stringstream s;
```
primaryodors, line 1006: Tempted to go back and fix this to create the string and have the operator<< stream this function, but perhaps there's a certain craftiness to doing it this way.

src/classes/constants.h
```diff
@@ -88,6 +94,30 @@
...
+#define allow_auto_hydroxy 0
+#define prerot_sidechains_from_ligand 0
+
+#define _DBG_LONELINESS 0
+#define _DBG_STEPBYSTEP 0
+#define _DBG_TUMBLE_SPHERES 0
+#define debug_break_on_move 0
+#define debug_stop_after_tumble_sphere 0
+#define _DORESPHRES 0
+#define include_old_intermol_conforms 0
```
primaryodors, line 119: #91.

src/classes/molecule.cpp
```diff
@@ -225,6 +260,25 @@ Atom* Molecule::add_atom(char const* elemsym, char const* aname, Atom* bondto, c
     }
 }
 
+void Molecule::save_state()
```
primaryodors, line 263: Get rid of these two functions. Anything that depends on them should use Pose objects instead.

src/classes/molecule.cpp
```diff
@@ -1217,10 +1271,20 @@ Bond** Molecule::get_rotatable_bonds()
...
+                bool pia = lb[j]->atom->is_pi(),
+                	 pib = lb[j]->btom->is_pi();
+                
+                int fa = lb[j]->atom->get_family(),
+                	fb = lb[j]->btom->get_family();
+                
+                // Generally, a single bond between pi atoms, or a bond from a pi atom to an amino group, cannot rotate.
+                if (	(pia && pib)
+                		||
+                		(pia && (fb == PNICTOGEN))
```
primaryodors, line 1285: This code exists, but its intended functionality is not working.

Check the AminoAcid class see if maybe something in there is preventing the CONH2 bond from being rigid. #89

src/classes/molecule.cpp
```diff
@@ -2622,7 +2765,19 @@ void Molecule::multimol_conform(Molecule** mm, int iters, void (*cb)(int))
                                 for (j=0; mm[j]; j++)
                                 {
                                     if (!nearby[j]) continue;
-                                    float lbind1 = mm[i]->get_intermol_binding(mm[j]);
+                                    float lbind1 = 
+                                    
+                                    #if allow_ligand_esp
```
primaryodors, line 2770: This is hard to read and could be made clearer, if not with a rewrite, then with a comment.

src/classes/protein.cpp
```diff
@@ -176,6 +176,24 @@ bool Protein::add_sequence(const char* lsequence)
     return true;
 }
 
+void Protein::save_state()
```
primaryodors, line 179: Can use these two functions for #92.

src/podock.cpp
```diff
@@ -736,30 +842,45 @@ int main(int argc, char** argv)
...
-#if _DBG_STEPBYSTEP
+			#if _DBG_STEPBYSTEP
             if (debug) *debug << "Saved last nodecen." << endl;
-#endif
+			#endif
 
             // Move the ligand to the new node center.
-            m.recenter(nodecen);
-#if _DBG_STEPBYSTEP
+            // m.recenter(nodecen);
```
primaryodors, line 857: Uncomment this and put it in an `#if !allow drift` block.

src/podock.cpp
```diff
@@ -736,30 +842,45 @@ int main(int argc, char** argv)
...
+			#if _DBG_STEPBYSTEP
             if (debug) *debug << "Conformer momenta reset." << endl;
-#endif
+			#endif
 
 
             sphres = p.get_residues_can_clash_ligand(reaches_spheroid[nodeno], &m, nodecen, size, mcoord_resno);
             for (i=sphres; i<SPHREACH_MAX; i++) reaches_spheroid[nodeno][i] = NULL;
+            
+            for (i=0; i<sphres; i++)
```
primaryodors, line 870: Add a comment here to indicate what this block does.

primaryodors (approved):
No show stoppers found, but check back there's some minor changes to make yet still. Can do them in the same branch and create a new PR when ready.

PR #96, opened 2022-07-26
Ssepeq pose class

ssepeq:


ssepeq:
The local server keeps intermittently segfaulting.

ssepeq:
cis-3-hexen-1-ol is now intermittently failing in OR1A1.

ssepeq:
Lyral and linalool have been failing in OR1A1. Lyral is not an agonist and is physically too big to fit in the binding pocket. Linalool is a weak agonist.

Saito et al (https://doi.org/10.1126/scisignal.2000016) report that lyral is a non-agonist, but Adipietro et al (https://doi.org/10.1371/journal.pgen.1002821) report it as an inverse agonist. Looking at the structure of lyral, both the 3D structure on pubchem and the structure generated by obabel have the aldehyde group pointing in the polar direction off the cyclohexene ring. Since PrimaryDock cannot yet bend that ring, it cannot move that ring member such that the aldehyde falls equatorially; if it could, the dock might succeed. 

Alternatively, there's no enantiomer selection in the SMILES string for lyral, and commercial samples of the odorant are probably typically racemic, it being a synthetic compound. (One manufacturer calls it a "mixture of isomers".) So it may be that one enantiomer is an inverse agonist while the other does not fit into the pocket. The 3D structure on Good Scents (http://www.thegoodscentscompany.com/opl/31906-04-4.html) has the aldehyde group positioned equatorially.

ssepeq:
Linalool still fails in OR1A1. It's worse at 50 iterations than at 100. 200 iterations takes way too long.

ssepeq:
200 iterations takes half an hour to generate one pose. Inexcusable.

ssepeq:
Waiting to see how many dismal failures this foolish thing generates on the local server before green lighting this for review and merge.

src/primarydock.cpp
```diff
@@ -579,7 +579,7 @@ int main(int argc, char** argv)
     {
     	ligand->minimize_internal_clashes();
     	float lig_min_int_clsh = m.get_internal_clashes();
-    	ligand->crumple(fiftyseventh*15);    	
+    	ligand->crumple(fiftyseventh*44);
```
ssepeq, line 582: Dratbil.

primaryodors (approved):

primaryodors (changes requested):
Unacceptable.

PR #100, opened 2022-08-01
Monte Carlo options for multimol conformer search.

ssepeq:


src/classes/molecule.cpp
```diff
@@ -2181,7 +2181,14 @@ void Molecule::multimol_conform(Molecule** mm, int iters, void (*cb)(int))
                 }
                 else
                 {
+                	#if monte_carlo_axial
```
primaryodors, line 2184: No rush, but eventually get around to changing these spaces to tabs. They're all over the changes made to the function.

primaryodors (approved):

PR #106, opened 2022-08-02
Test config files for known receptor-ligand pairs.

ssepeq:
OR1A1 d-limonene test.


ssepeq:
The butyrate one is doing something weird...

```
PDB file: pdbs/OR1/OR1A1.rotated.pdb
Ligand: sdf/butyrate.sdf


5 pose(s) found.

Calculation took: 4 seconds.
```

The ligand is an inverse agonist for the receptor, so it should dock inside the pocket in a way that does not cause receptor activation. However, the docker seems to just give up without even trying to dock it. Then it reports 5 poses (the maximum configured in the input file) but no actual PDB data on the screen or in the output file.

ssepeq:
`make clean` fixed it.

ssepeq:
The acetophenone one is still doing the strange behavior. This ligand-receptor pair is a non-agonist, however the docker should still return either a statement saying 0 poses or else PDB output for any poses that it does find.

ssepeq:
Marking this ready since it contains a code fix for the docker.

primaryodors (approved):

PR #111, opened 2022-08-04
Differential energy computation.

ssepeq:
Output of values for before vs. after ligand docking, and delta values.


src/classes/constants.h
```diff
@@ -106,6 +106,9 @@
 #define monte_carlo_axial 0
 #define monte_carlo_flex 1
 #define pocketcen_is_loneliest 1
+#define pre_ligand_multimol_radius 15
```
primaryodors, line 109: Consider making some or all these three items variables dependent on Point size.magnitude() in primarydock.cpp main().

src/classes/intera.cpp
```diff
@@ -443,6 +443,8 @@ InteratomicForce** InteratomicForce::get_applicable(Atom* a, Atom* b)
 
             case hbond:
                 // if (sgn(a->is_polar()) == -sgn(b->is_polar()))
+                if (a->get_family() == PNICTOGEN && a->is_amide()) break;
```
primaryodors, line 446: Add a check for is_backbone() for better performance.

src/classes/molecule.cpp
```diff
@@ -1944,7 +1944,8 @@ float Molecule::get_intermol_binding(Molecule** ligands)
     if (!ligands[0]) return 0;
     int i, j, l;
     float kJmol = 0;
-    kJmol -= get_internal_clashes();
+    if (!atoms) return 0;
+    if (!atoms[0]->residue) kJmol -= get_internal_clashes();
```
primaryodors, line 1948: This isn't quite right. Instead of obliterating all internal clash calculations altogether for residues, code should ensure the Molecule::base_internal_clashes property is set for all residue AAs.

Since the looped calls on lines 649 and 1412 of primarydock.cpp call this function repeatedly, it may be helpful to add an optional boolean parameter something like subtract_clashes = true, and in the loop pass it in as j==0.

src/classes/molecule.cpp
```diff
@@ -1959,6 +1960,7 @@ float Molecule::get_intermol_binding(Molecule** ligands)
             if (ligands[l] == this) continue;
             for (j=0; j<ligands[l]->atcount; j++)
             {
+            	if (atoms[i]->is_backbone && ligands[l]->atoms[j]->is_backbone) continue;			// kludge to prevent adjacent residue false clashes.
```
primaryodors, line 1963: This should be expanded to check that the residue numbers differ by 1, the names of the atoms are C and N, and that the C is the lower numbered residue.

src/classes/molecule.cpp
```diff
@@ -2040,17 +2043,47 @@ void Molecule::minimize_internal_clashes()
...
     for (i=0; mm[i]; i++)
     {
         mm[i]->reset_conformer_momenta();
     }
     inplen = i;
+    
+    if (!bkg) bklen = 0;
+    else
+    {
+		for (i=0; bkg[i]; i++);		// Get count.
```
primaryodors, line 2066: Indentation.

src/primarydock.cpp
```diff
@@ -588,7 +599,67 @@ int main(int argc, char** argv)
 			p.load_pdb(pf);
 			fclose(pf);
     	}
-    	else
+    	
+    	if (differential_dock)
+    	{
+			std::vector<AminoAcid*> preres = p.get_residues_near(pocketcen, pre_ligand_multimol_radius);
```
primaryodors, line 605: Indentation.

src/primarydock.cpp
```diff
@@ -1249,11 +1320,25 @@ int main(int argc, char** argv)
...
 
             // time_t preiter = time(NULL);
-            Molecule::multimol_conform(
-                cfmols,
-                iters,
-                &iteration_callback
-            );
+            if (differential_dock)
+            {
+		        Molecule** delete_me;
```
primaryodors, line 1325: Indentation.

src/primarydock.cpp
```diff
@@ -1286,10 +1373,50 @@ int main(int argc, char** argv)
...
+                imvdWrepl[metcount] = 0;							// TODO
+                
+                metcount++;
+                
                 btot += lb;
                 // cout << "Metal adds " << lb << " to btot, making " << btot << endl;
             }
+            
+            float final_binding[seql+4];
+			float final_vdWrepl[seql+4];
```
primaryodors, line 1390: Indentation, all the way to line 1419.

src/primarydock.cpp
```diff
@@ -1298,12 +1425,41 @@ int main(int argc, char** argv)
                 if (!reaches_spheroid[nodeno][i]) continue;
                 if (!p.aa_ptr_in_range(reaches_spheroid[nodeno][i])) continue;
                 reaches_spheroid[nodeno][i]->clear_atom_binding_energies();
-                float lb = m.get_intermol_binding(reaches_spheroid[nodeno][i]);
+                int resno = reaches_spheroid[nodeno][i]->get_residue_no();
+                
+                float lb;
+                if (differential_dock)
+            	{
```
primaryodors, line 1432: Indentation, this and the next block. What is wrong with me? I used to be so good at indenting properly. Or believed I was.

src/primarydock.cpp
```diff
@@ -1298,12 +1425,41 @@ int main(int argc, char** argv)
...
-                sprintf(metrics[metcount], "%s%d", reaches_spheroid[nodeno][i]->get_3letter(), reaches_spheroid[nodeno][i]->get_residue_no());
+                
+                sprintf(metrics[metcount], "%s%d", reaches_spheroid[nodeno][i]->get_3letter(), resno);
                 // cout << metrics[metcount] << ": " << lb << " . ";
                 mkJmol[metcount] = lb;
-                mvdWrepl[metcount++] = m.get_vdW_repulsion(reaches_spheroid[nodeno][i]);		// TODO: Include repulsions between side chains.
+                imkJmol[metcount] = initial_binding[resno];
+                
+                if (differential_dock)
+            	{
```
primaryodors, line 1447: Indentation.

src/primarydock.cpp
```diff
@@ -1437,7 +1601,15 @@ int main(int argc, char** argv)
                         cout << "Pose: " << i << endl << "Node: " << k << endl;
                         if (output) *output << "Pose: " << i << endl << "Node: " << k << endl;
 
-                        cout << "# Binding energies" << endl << "BENERG:" << endl;
+                        if (differential_dock)
+				    	{
```
primaryodors, line 1605: May want to just use astyle to convert all tabs to spaces and then recheck the entire file.

primaryodors (changes requested):
A few things to address before merge.

src/classes/molecule.cpp
ssepeq: Doing it that way still produces occasional unrealistically large numbers. Kept the subtract_clashes param, but it's currently set to always false for these specific calls.

primaryodors (approved):

PR #116, opened 2022-08-08
Efforts towards a prediction algorithm.

ssepeq:
Writing the tools that will hopefully be able to predict the activity of odorants in receptors.


ssepeq:
There's stuff in here that shouldn't have to wait for good predictions (as well as in progress stuff that won't hurt anything).

primaryodors (approved):

PR #117, opened 2022-08-09
Efforts towards a prediction algorithm, part II.

ssepeq:
Writing the tools that will hopefully be able to predict the activity of odorants in receptors.


ssepeq:
So far, it seems there is not a correlation between known agonists and non agonists vs. their inactive and active docked energies. Both produce lower energy levels in the inactive configuration, whether differential or not. Further, even known agonists that should easily fit into the binding pocket are generating few decent (i.e. non clashing) poses, and the poses that are generated do not have good polar satisfaction.

Fixing the pose generation might improve the correlations.

ssepeq:
To effectively design the docking paths for receptors, it would be best to scour the literature for all mutations known to alter receptor function. A great example is https://doi.org/10.1021/jacs.6b06983 which gives seven sites for OR2T11 that, if mutated, result in zero response to the receptor's strongest odorant, as well as several other sites that have either no effect or a lesser effect if mutated.

Which sites are important to ligand binding may vary greatly from one receptor to another, and the prediction code might have to examine protein sequences for specific features that indicate which of several binding sites to employ.

ssepeq:
OR1A1 G108, G152, I205. https://doi.org/10.1007/s00018-017-2576-z; Y251, Y258. https://doi.org/10.1073/pnas.1713026115
OR2T11 M56, M115, M133, R135, C138, C238, H241. https://doi.org/10.1021/jacs.6b06983
OR5AN1 Y260. https://doi.org/10.1073/pnas.1713026115

ssepeq:
A different source for PDB files of the receptors:

https://alphafold.ebi.ac.uk/entry/Q8NHB7

Might be more accurate than ZhangLab? The HXXCD motif still isn't helixed, but it's right off the edge of a helix and is positioned down near the binding pocket, meaning less change would be necessary to metal-bind it.

ssepeq:
AlphaFold's models lack hydrogens.

ssepeq:
Getting nothing but wrong results from this test:

```
clear; make primarydock && php -f predict/dock.php prot=TAAR8 lig=cadaverine
```

Somehow, the activation matrix seems to be pulling TMR3:Asp111 *away* from the ligand bound to TMR5:Asp201, when in fact the extracellular ends of TMR3 and TMR5 should be moving closer together. The bad conformational change of the TMRs results in a consistently incorrect prediction of either inverse agonist due to increased total energy, or non-agonist due to the lack of an active node result within energy limits.

```
    "TAAR8":
    {
        "cadaverine":
        {
            "Inactive": -37.6059,
            "Active": -22.163917499999997,
            "Prediction": "Inverse Agonist",
            "Actual": "Agonist"
        }
    }
```

ssepeq:
The predictions that it's generating are not right, but nevertheless the code and infrastructure are in place to make predictions. Getting the matrices right will allow predictions with at least 60% accuracy for most ORs, and up to 99% accuracy for a very few ORs. Also keep in mind that the source data are probably only about 80% accurate at best, not counting Mainland et al (2015) which seems to be less accurate (though how much less is not known).

After 21 days and 73 commits, it's time to put this one up for merge.

primaryodors:
Big Three tests so far: 🟩🟩

primaryodors (approved):

PR #123, opened 2022-08-21
Multiple scripting commands including REGION and CENTER.

ssepeq:


primaryodors:
Performance has really taken a serious hit! The TAAR test just took 589 seconds - it used to take a quarter of that.

primaryodors (approved):

PR #145, opened 2022-11-03
Water in the binding pocket.

ssepeq:
This is working now and should be good to merge.


src/primarydock.cpp
```diff
@@ -1031,6 +1078,19 @@ int main(int argc, char** argv)
     pocketcen = loneliest;
     #endif
 
+    if (waters)
+    {
+        float szscale = 0.5;
+        for (i=0; i<maxh2o; i++)
+        {
+            waters[i]->recenter(Point(
+                pocketcen.x + frand(-size.x * szscale, size.x * szscale),
```
primaryodors, line 1087: One feature that would be nice to add is in a path based dock, to move all the water molecules into the new node if it is more than a threshold distance from the previous node, and then settle them in so they aren't clashing.

primaryodors (approved):

PR #155, opened 2022-11-20
Odorant page: list indicates receptors w/o sufficient data.

ssepeq:
Receptors with less than two ligands of any activity level, or without any aroma note correlations meeting the cutoff threshold, will show as (insufficient data) in dim italic text instead of repeating notes from the current compound or just saying "none".

No changes to primarydock code.

### DIFF
--- a/www/receptor_notes.php
+++ b/www/receptor_notes.php
@@ -22,8 +22,8 @@ function correlate_receptors_aromanotes()
     foreach (array_keys($prots) as $rcpid)
     {
         $ep = all_empirical_pairs_for_receptor($rcpid, true);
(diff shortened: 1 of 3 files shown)


www/receptor_notes.php
```diff
@@ -21,6 +21,10 @@ function correlate_receptors_aromanotes()
 
     foreach (array_keys($prots) as $rcpid)
     {
+        $ep = all_empirical_pairs_for_receptor($rcpid, true);
+        // foreach ($ep as $k => $v) if ($v <= 0) unset($ep[$k]);
```
ssepeq, line 25: Leaving this here in case we want it later.

Issue #149, opened 2022-11-09
Viewer: Visually indicate the name of a loaded protein model.


ssepeq:
When loading a PDB or .dock file with the choose file button, ensure that the name of the protein or protein+ligand is/are displayed somewhere on the screen.

primaryodors:
#156 does not fix for .dock files, only for PDBs.

PR #156, opened 2022-11-25
Metal bindings of ORs.

ssepeq:
Can now use `predict/metal.php` to create metal-bound versions of all ORs that have metal binding sites.

The metal coordination feature has been significantly updated and gives better results in the positions of the side chains.

The prediction system has been upgraded, specifically `predict/method_combined.php`, and the algorithms in `predict/dock_eval.php`.

There is also a viewer fix for #149.


primaryodors:
🟩🟩🟩

PR #161, opened 2022-11-29
Pnictogen protonation.

ssepeq:
Originally this was a trip switch feature to measure and output clashes with a sequence of trip switch residues. However, it seems ORs do not work that way. This branch retains trip switch code albeit without documentation, and also includes some viewer enhancements and code to make acids and neutral pnictogens interact as though the pnictogen had up to a 25% partial protonation.


primaryodors:
🟩🟩🟩

PR #165, opened 2022-12-01
Mixtures feature.

ssepeq:
Users can now mix odorants and see an estimate of which receptors will be activated and by how much.

No changes to docking code.


primaryodors:
This is not how dose response curves work.

PR #166, opened 2022-12-01
Extrapolations for when only top or only ec50 known but both required.

ssepeq:
Based on a scatter plot of curve tops (x axis) vs. ec50 values (y axis):

image

The red line equals the function: `$est_ec50 = -$top / 5.3 - 3.7;`.


ssepeq:
Not an official PrimaryDock Web App code file, but this is what generated the above image:

[top_vs_ec50.php.txt](https://github.com/primaryodors/primarydock/files/10129302/top_vs_ec50.php.txt)

PR #168, opened 2022-12-03
Revert "Extrapolations for when only top or only ec50 known but both required."

ssepeq:
Reverts primaryodors/primarydock #166

Go to the page for OR2W1 and scroll all the way to the bottom. The calculation is mixing agonists with inverse agonists.


primaryodors:
Probably easier to just create a new branch and a new PR with the fix.

PR #169, opened 2022-12-04
Limit shielding allowances for pi bonds.

ssepeq:
Currently, a potential interatomic interaction is considered shielded if the angle between the two atoms, relative to the shielding atom is >= 130deg. This PR has the effect of reducing this threshold to 100deg for pi bonds.

The shielded status is currently unfortunately all-or-nothing, with no accounting for partial shielding caused by partial eclipses of atoms. While this PR does not address that shortcoming, it does provide a new public Atom::shielding_angle property that contains the result of the most recent shielding check, that can be used inside the interaction class for partial shielding calculations.

### DIFF
--- a/src/classes/atom.h
+++ b/src/classes/atom.h
@@ -262,6 +262,7 @@ public:
     float last_bind_energy = 0;
     float strongest_bind_energy = 0;
     Atom* strongest_bind_atom = nullptr;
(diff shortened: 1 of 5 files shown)


primaryodors:
🟩🟩🟩

PR #170, opened 2022-12-04
Top/ec50 fix.

ssepeq:
Fixes the sort order in the ligands list of the receptor page and the receptors list of the odorant page.

Fixes the mixtures feature to use a proper dose response curve sigmoid formula.

Calibrated the mixtures feature using floralozone dil. in EtOH by comparing a pink-floral-jasmine note of the mixture with the estimated OR5B17 activation level at each dilution. The numbers given will represent the maximum odor strength from sniffing directly above the lip of a 10ml vial.


ssepeq:
No change to the docker or its classes.

Issue #171, opened 2022-12-04
Mixture page is still not calculating properly.


primaryodors:
http://www.primaryodors.org/mixture.php?mixttl=%2B2J2+-6P1&9b50b0710bce756d895b0a704720c2f6=1&8182e49cd4e0126f80b62cde4815d47c=1&47029b0a89b677c61d67c3a1b1071f07=1500&add=ee4dd2290d52c459c35d7b6c3203f231&addwt=1

OR1A1 should be a little more active than OR1A2 in this mixture, due to the octanol, but in fact it shows up much farther down the list than OR1A2 does, and fails to list 1-octanol among its agonists.

PR #172, opened 2022-12-04
Fix for mixture page and odorant list on receptor page. #171

ssepeq:


ssepeq:
No changes to docking code.

Issue #146, opened 2022-11-03
Move water molecules in a path based dock.


primaryodors:
One feature that would be good to add is in a path based dock, to move all the water molecules into the new node if it is more than a threshold distance from the previous node, and then settle them in so they aren't clashing.

If the new node is near enough the old node that their water spaces overlap, any water molecules inside that overlap should stay put and water molecules outside the new water space should be moved into a non-overlapping part of the new water space.

_Originally posted by @primaryodors in https://github.com/primaryodors/primarydock/pull/145#discussion_r1012497655_

ssepeq:
This is partially addressed by: https://github.com/primaryodors/primarydock/commit/52e582b9f128482724cf0a97109bb54ffad168a5

However, we still want to add distance to the ligand target center as a reason to teleport, not just clashes and polar dissatisfaction.

ssepeq:
Still happening: once a water molecule is deleted, it is unavailable for future poses. We should bring back deleted waters when a new pose begins.

PR #175, opened 2022-12-12
Multiple enhancements.

ssepeq:
- Config files for docking eugenol in all of the OR10G proteins;
- When config args are passed on the command line, echo them back to confirm;
- Dissatisfied water molecules teleport or are deleted;
- Histidine hydrogen flip. #8
- Move water molecules with path node, bringing back any that were deleted. #146
- Web app: Correct spatial arrangement of residues in odd numbered TM helices;
- Ability to specify best-binding or tumble spheres modes from the config file and the command line;
- Viewer: Show BW50 numbers in the side chain show/hide popup after loading a dock;
- Output poses to always be numbered sequentially.


ssepeq:
For OR10G7-eugenol I am getting really poor output as far as number of poses; with a max set of 20, it is giving me only 3 good ones. This is for a mutually strongest agonist pair, i.e. the best known agonist for the receptor and the best known receptor for the odorant.

primaryodors:
🟩🟩🟩

Issue #158, opened 2022-11-26
Move data files into the data folder.


ssepeq:
Currently, we have data files in the project root folder that would be better placed under data/, namely:

- aminos.dat
- bindings.dat
- elements.dat

This also involves repointing all code that uses these data files.

PR #179, opened 2022-12-16
Move data files to data folder.

ssepeq:
See #158.

Additionally, amino acid 3-letter codes are now case insensitive.

No changes to the code that does the intermolecular binding, but might be a good idea to retest Big 3 anyway.


ssepeq:
🟩🟩

ssepeq:
It occurs to me that vdW-binding atoms should not be paired with hydrophilic amino acids:

```
# Best binding heavy atom 2 of ligand
# LBBA: C1 type: vdW

...

# Aligning C1 to Asn109...
```

I should like to include a fix for that in this PR.

ssepeq:
🟩🟩🟩

PR #180, opened 2022-12-21
Maintain locks and zippers from original PDB.

ssepeq:
Docking will no longer break apart any salt bridges, leucine zippers, or other locks in place unless the ligand would provide a better intermolecular bond.

### DIFF
--- a/src/classes/constants.h
+++ b/src/classes/constants.h
@@ -177,7 +177,7 @@
 #define default_bestbind 1
 #define preemptively_minimize_intermol_clashes 0
 #define allow_bestbind_springiness 1
-#define bestbind_springiness 10
+#define bestbind_springiness 15
 
 // For differential docking, whether to multimol_conform() all the protein's residues into an
 // optimized initial conformation before adding the ligand.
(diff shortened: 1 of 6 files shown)


ssepeq:
Bug: The dock keeps bringing heavy atoms way too close together. This is especially causing the amino aldehyde test to seriously fail.

ssepeq:
🟩🟥🟩

ssepeq:
🟩🟨🟨

The TAAR8 and OR1A1 tests do work, but intermittently. The dock doesn't always find a good pose, likely because the ligand keeps ending up bumped up against the coordination residues.

ssepeq:
🟩🟩🟩

primaryodors (changes requested):
This PR does not accomplish its intended purpose: "Maintain locks and zippers from original PDB.".

image 
image 
image

The salt bridge between Asn111 and His243 in OR1A1 should remain like it is in the input PDB:
image

ssepeq:
🟩🟥🟩

ssepeq:
TAAR8 test is failing because of the Y116 - D201 hydrogen bond. Y116 is clashing with the ligand but not always producing a negative binding energy, therefore forbidding residue flexion. The docker compromises by moving the ligand out of the way, resulting in incorrect poses (and incidentally no poses within energy limits).

ssepeq:
Abandoning this.

PR #181, opened 2022-12-23
Avoid moving side chains that are strongly bound in the original PDB.

ssepeq:
Mark strongly bound (< -5 kJ/mol) residue side chains as MOV_NONE unless clashing with the ligand. Also partially charge neutral, non-amide nitrogen in proximity to a negative charge. Also fix springy bond distances.

Currently testing Big 3. The OR1A1 test should reveal whether Asp111 stays salt bridged to His243.

### DIFF
--- a/src/classes/protein.cpp
+++ b/src/classes/protein.cpp
@@ -222,6 +222,53 @@ void Protein::end_pdb(FILE* os)
     fprintf(os, "END\n");
 }
 
+void Protein::find_residue_initial_bindings()
+{
+    if (!residues) return;
+    
+    int i, j, k;
+    for (i=0; residues[i]; i++)
+    {
+        AminoAcid** aa = residues; // get_residues_can_clash(residues[i]->get_residue_no());
+
+        // If the current residue has one or more negatively charged non-backbone atoms,
+        // and a nearby residue has one or more neutral non-backbone pnictogen atoms not part of an amide,
+        // put a positive charge on the pnictogen(s).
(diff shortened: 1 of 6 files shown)


ssepeq:
🟩🟩🟩

primaryodors:
Another approach to consider might be only allowing the side chains to flex if both members of the lock are part of the Molecule** param that's passed into multimol_conform().

ssepeq:
🟩🟩🟩

src/classes/protein.cpp
```diff
@@ -222,6 +222,53 @@ void Protein::end_pdb(FILE* os)
...
+    
+    int i, j, k;
+    for (i=0; residues[i]; i++)
+    {
+        AminoAcid** aa = residues; // get_residues_can_clash(residues[i]->get_residue_no());
+
+        // If the current residue has one or more negatively charged non-backbone atoms,
+        // and a nearby residue has one or more neutral non-backbone pnictogen atoms not part of an amide,
+        // put a positive charge on the pnictogen(s).
+        if (residues[i]->get_charge() < -0.5)
```
primaryodors, line 237: The effect of this will be to put a charge on *all* histidines. Since it isn't filtering by proximity to an acid, the presence of acidic side chains anywhere in the protein will have this undesired effect.

primaryodors (changes requested):
Once the histidine charge effect is fixed, this will be ready to merge.

primaryodors (approved):
This should be fine.

PR #184, opened 2022-12-24
Fix hard coded glycine check in primarydock.cpp

ssepeq:
Added `bool AminoAcid::is_glycine()` to check if the object is a glycine molecule or residue, obviating the hard coded caseless `name == 'Gly'` kludge that was in the primarydock app.

As this affects the best-binding algorithm, it should be tested on multiple ligands with aliphatic groups.


ssepeq:
🟩🟩🟩

PR #188, opened 2022-12-28
Ability to choose between PDB sources.

ssepeq:
This PR includes:

- Ability to choose AlphaFold or ZhangLab as sources for receptor PDBs;
- Default receptor PDBs are now from AlphaFold;
- Ability to hydrogenate PDBs in the editor;
- Complete information about OR9G9;
- Viewer is able to read PDBs from RCSB and display ligands.


ssepeq:
🟩🟥🟩

Looking into why d-limonene is failing in OR1A1...

ssepeq:
Isoleucine is having a similar problem to what threonine was doing; the hydrogens on the delta carbon are not bonding to the carbon, causing them to get left behind when the side chain flexes.

The 1A1 test actually intermittently works, despite the Ile bug. 🟩🟩🟨

ssepeq:
🟩🟩🟩

Note that the AlphaFold version of OR1A1 does not have the Asp111 - His243 salt bridge, so this part of the test doesn't apply.

src/classes/atom.h
```diff
@@ -317,6 +317,8 @@ class Atom
     static float atomic_weights[_ATOM_Z_LIMIT];
     static int valences[_ATOM_Z_LIMIT];
     static int geometries[_ATOM_Z_LIMIT];
+
+    friend class AminoAcid;
```
primaryodors, line 321: Why is AminoAcid a friend class?

src/classes/molecule.cpp
```diff
@@ -303,11 +303,22 @@ Atom* Molecule::add_atom(char const* elemsym, char const* aname, Atom* bondto, c
 
 void Molecule::clear_all_bond_caches()
 {
-    Bond** b = get_rotatable_bonds();
+    /*Bond** b = get_rotatable_bonds();
```
primaryodors, line 306: Can delete this if it is dead code.

src/classes/molecule.cpp
```diff
@@ -358,6 +370,50 @@ void Molecule::hydrogenate(bool steric_only)
...
 
+        int fam = atoms[i]->get_family();
+        if (fam == PNICTOGEN || fam == CHALCOGEN)
+        {
+            Atom* C = atoms[i]->is_bonded_to_pi(TETREL, true);
+            if (!C) C = atoms[i]->is_bonded_to_pi(PNICTOGEN, true);
+            if (C)
+            {
+                atoms[i]->aromatize();
+                // TODO: Rotate atoms[i] geometry to coplanar with pi bond.
```
primaryodors, line 381: The functionality has been added to rotate the geometry? This still says TODO.

primaryodors (changes requested):
Ok to merge this once the changes have been addressed. 👍🏼

src/classes/atom.h
ssepeq: Oops!

primaryodors (approved):

PR #190, opened 2023-01-02
Ability to create disulfide bonds in the editor.

ssepeq:
A new `DISULF` command has been added to pepteditor that creates disulfide cross links between cysteine and/or homocysteine residues.

No change to docking functions.


ssepeq:
TODO: Read and interpret any `CONECT` records in the input PDB and store them in the vector<> that's used for outputting `CONECT` records.

ssepeq:
The AlphaFold models imply that the vast majority of the receptors with an HXXCD/E motif in the EXR2 domain cannot be helixed and metal bound, because the C in that motif would be disulfide bonded to a C near the top of TMR3, and there would not be space for an alpha helix in that section of the EXR2 domain. Even in receptors with the motif and no TMR3 disulfide site, the EXR2 domain offers no room for a helix.

At minumum, such a disulfide bond precludes the HXXCD/E motif from acting as a ligand capture device.

Issue #187, opened 2022-12-25
Test docker against X-ray diffraction PDBs.


ssepeq:
Here are several PDBs of the human adenosine receptor ADORA2 bound to various ligands, made using X-ray crystallography:

https://tinyurl.com/yc7bckxx (links to rcsb.org).

Have to come up with a way to either convert these into .dock format, or better yet modify the viewer to show the ligands in ball and stick format. Once this is done, the crystallography models can be used to verify the accuracy of primarydock and make any necessary improvements.

ssepeq:
ADORA2A has Phe168 in the EXR domain that in the X-ray models consistently pi-stacks with the bicyclic groups of both adenosine and caffeine. But the docker attempts to align the aromatic nitrogens with Glu13, Thr88, Asn181, Glu169, Asn253, etc. At least in the case of aromatic heterocycles, it seems pi stacking ought to have a stronger effect.

ssepeq:
With the most recent commits of the `ssepeq-acv1st` branch, the pi stacking is working flawlessly however there ought to be more ionic attraction in positioning the ligand, and perhaps more tendency to favor poses with pnictogen protonation.

ssepeq:
For the adenosine dock to work well, the functional groups of amino, aromatic nitrogenous polycycle, and polyol would have to each be considered as distinct units for the best-binding algorithm, in lieu of the current system that allows only a single heavy atom and optionally one of its hydrogens to be considered for each of the 3 positions. Then the amine could be aligned to Glu169, the heterocycles to Phe168, and the polyol to His278.

But since PrimaryDock is written mainly for olfactory reception, and adenosine is a heavier molecule than any odorant, the algorithm should be attuned to odorant ligands. Coplanar heterocycles such as pyrazines, indoles, and maltols would be capable of hydrogen bonding on their hetero atoms, while simultaneously pi stacking as conjugated rings. In these cases, the pi-stacking rings of pyrazine and indole would be considered the number 1 "most bindable" entity, and the hydrogen-bonding nitrogens/NH groups second, while the polar hydroxy group of maltol would be its first, followed by the pi ring and then the other two h-bonding oxygens. So certain atoms could be part of more than one functional group, even in cases where they may be the only atom in one of the groups.

ssepeq:
https://www.nature.com/articles/ncomms5202

[...]

Issue #191, opened 2023-01-03
Update functionality for tugs (related to #187)


ssepeq:
The ssepeq-acv1st branch is doing fairly well docking adenosine in ADORA2. The base is pi-stacked to Phe168, the sugar has two hydroxyls pointed towards His278, but the amino group is not as close to Glu169 as in the X-ray model.

Unmet potential binding between ligand and side chains should elicit a tugging effect on the ligand, and on specific atoms of the side chains. The forces involved for the ligand would be linear transformational, barycentric rotational, atom-centric rotational, and flexional; the forces for the side chain would be flexional only.

Anisotropy would slightly complicate the algorithm, specifically: Ligand atom a "wants" side chain atom b to be at location wb; side chain atom b "wants" movable atom a to be at location wa; flexionals to bring b closer to wb; flexionals to bring a closer to wa; barycentric rotational to bring a closer to wa; linear transformation to bring a closer to wa; atom-centric rotational to bring wb closer to b.

Each of these forces would have a partial effect each iteration, i.e. they would not serve to instantly bring each atom right where it is "wanted", but rather to move atoms closer to their target locations, by a factor of between 0.1 and 0.333 each iteration.

ssepeq:
The ssepeq-acv1st branch is giving inconsistent results docking adenosine in ADORA2, because of course it 🤬ing is.

ssepeq:
Tugs have been implemented in one of the stale branches. The result is a disaster.

primaryodors:
Closing this as tugs are unnecessary and in fact consistently produce worse results.

PR #192, opened 2023-01-08
Activate-First Docking

ssepeq:
This is a work in progress. The final PR will perform active vs. inactive calculations by docking the ligand in the active configuration first, and then computing the same pose in the inactive conf.

~Along the way, efforts are being made to improve docking for receptor-ligand pairs that have X-ray diffraction models available. See #187 for more info.~


ssepeq:
🟥🟥 ...

The amino aldehyde test reports a binding energy of about -19, however the two amino groups are too close together and pointing hydrogens at each other, which is incorrect because the hydrogens' polarities would repel.

TAAR test only gives one result and there's not enough ionic attraction.

ssepeq:
Something to try might be instead of tugs (see #191), to have a list of potential interactions in the multimol_conform() function, filtered so that only atoms within a threshold distance would count, and probably only within a threshold angle relative to the ligand's barycenter. Such unsatisfied interactions would at minimum affect the scoring of the pose candidates, and may even exert translational and rotational effects on the existing whole molecule momenta.

ssepeq:
🟩🟩 ...

ssepeq:
🟥🟩🟩

ssepeq:
🟩🟩🟩

ssepeq:
It might be a while before this one is ready.

PR #203, opened 2023-03-04
Hypothesized notes for receptors.

ssepeq:
This PR adds hypothesized aroma qualities that each receptor is thought to contribute to odor perception. The hypothesizeds appear in a new column in odorant view on the web app.

Also included is a fix for the molecular structure of butyl anthranilate.


ssepeq:
No change to C++ code, no necessity to run Big Three.

PR #205, opened 2023-03-04
Arginine fix from #202.

ssepeq:
- Arginine no longer allows the mistaken hydrogen flip. Only uncharged HN-C=N groups (and congeners) are allowed.
- All pnictogens bound to pi tetrels are themselves pi, unless also bound to a chalcogen.


ssepeq:
🟩🟩...

ssepeq:
Histidine flips still work. Merging this in.

PR #206, opened 2023-03-04
Ligand bias and centralize multimol calculation from #202.

ssepeq:
- There's a new constant called `dock_ligand_bias` that represents how much proportional additional strength is computed for ligand-protein bindings during multimol dock;
-  The multimol trial-and-error comparison calculation has been moved to a new function called `Molecule::intermol_bind_for_multimol_dock()`.


ssepeq:
🟩🟩🟩

PR #207, opened 2023-03-04
Polar satisfaction from #202.

ssepeq:
- Polar satisfaction is now factored into the best-binding algorithm, the dock algorithm, and the scoring function.
- All three places for polsat influence can be independently adjusted in constants.h.


ssepeq:
🟩🟥🟩

ssepeq:
Without changing anything, it suddenly worked.

🟩🟩🟩

PR #208, opened 2023-03-05
Upright the PDBs.

ssepeq:
Replaced incorrect modulus operator with bitwise and.


ssepeq:
No change to classes/docking code.

PR #209, opened 2023-03-12
"Soft-rocking" activation method.

ssepeq:
Per the recent findings by Billesbolle et al, a prediction method now exists that performs a rocking motion of TMR6 so that its extracellular end moves towards TMR4 and its cytoplasmic end moves out from the center of the protein.

Also included is an internal clashes calculation of the Protein class, with a repacking option so that the helix can move without creating avoidable clashes.

Also included is a bool in the Molecule class that stores whether a residue has been flexed, so that the output file will contain not just ligand-adjacent residues but also any residues flexed as part of internal clash avoidance repacking.

Also included in the "soft rock" correlation script is a fix for the colors of scatter plots.


ssepeq:
🟩🟩🟩

ssepeq:
The "soft rock" method does not make good predictions. It always rocks TMR6 whether or not the active configuration has a more favorable energy. Further, no meaningful correlation is observed between empirical agonism and bonding to TMR4, TMR5, TMR6, or the product of TMR4*TMR6 bindings. More work is necessary in order to produce useful predictions.

ssepeq:
🟩🟥

ssepeq:
🟩🟩🟩

PR #210, opened 2023-03-13
Hydrophobic cohesion.

ssepeq:
A.k.a. oil cuddling. Hydrophobic atoms will now exhibit a strong tendency to stick together, since their surroundings will be assumed to be predominantly hydrophilic as they are in olfactory receptor EXR and CYT regions, and in the spaces between the TM helices.

A new constant named `oxytocin` mediates the effect, in order to balance the amino aldehyde test, which prefers low values of the new constant, and the following dialkane test, which prefers high values:

`clear; make test/molecule_test && ./test/molecule_test 'CCCCCC' 'CCCCCC'`

A successful dialkane result shows the two chains lined up side by side, e.g.:

image

### DIFF
--- a/src/classes/constants.h
+++ b/src/classes/constants.h
@@ -292,7 +292,7 @@
 #define _dbg_glomsel 0
 #define _dbg_polsat 0
 #define _dbg_softrock 0
-#define _dbg_rock_pic 1
+#define _dbg_rock_pic 0
 
 #endif
 
--- a/src/classes/constants.h
+++ b/src/classes/constants.h
@@ -12,6 +12,7 @@
 #define _kcal_per_kJ 0.239006
 #define _kJmol_cuA 0.5
 #define vdw_clash_allowance 1.0
+#define oxytocin 0.007
 #define _DEFAULT_INTERA_R_CUTOFF 8
 #define _INTER_TYPES_LIMIT 10
 #define BOND_DEF_NOT_FOUND 0xbadb09d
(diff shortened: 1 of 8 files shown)


ssepeq:
TODO:

- Currently the cohesion effect affects all atoms. It should be modified to affect only nonpolar atoms.
- Currently the effect is "always-on". It would be best to add a boolean that the effect can be turned on and off.
- Even better would be the ability to define spherical and cylindrical regions where the effect takes place.

ssepeq:
🟩🟩🟨

ssepeq:
🟩🟩...

ssepeq:
🟩🟩...

ssepeq:
I am going to think more on the inclusion and exclusion zones. For now, here's the simple version of the feature.

primaryodors:
Should enable the new feature for the TAAR8 and OR1A1 tests.

ssepeq:
🟩🟩🟩

primaryodors:
Checking this out with propionate (strong agonist), hexanoate (non-agonist), and octanoate (non-agonist) in OR51E2.

image
image
image

Both the C6 and C8 acids have a tendency to orient themselves in a more vertical/diagonal position, which would draw the TM6 helix in closer. Predicting these protein-ligand pairs might involve guided docking (a feature planned but not yet written), and might not reveal a more favorable energy in the inactive state, but rather any gain in energy for the active state might fail to meet a certain threshold value determined by the protein structure.

PR #213, opened 2023-03-14
Recalculate can-clash residues when move part of a protein.

ssepeq:
`Protein::set_clashables()` now accepts an optional int resno. When `iteration_callback()` calls `protein->get_internal_clashes()` with a start and end resno, it will recompute the nearby clashable residues for the segment of protein in motion.

This does, however, prevent the soft rock activation of 51E2 by propionate.

TODO: `Protein::set_clashables()` should also recompute the clashables for each newly identified nearby residue. If done as a recursion, it should be limited to one layer deep.

### DIFF
--- a/src/classes/protein.cpp
+++ b/src/classes/protein.cpp
@@ -279,6 +279,7 @@ float Protein::get_internal_clashes(int sr, int er, bool repack)
 
         if (repack)
         {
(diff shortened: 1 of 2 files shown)


ssepeq:
Duplicate of a change made for #214 .

Issue #200, opened 2023-02-25
An OR has been directly observed in its agonist bound state.


primaryodors:
https://doi.org/10.1101/2022.12.20.520951
https://www.biorxiv.org/content/10.1101/2022.12.20.520951v1.full.pdf

Protein structure (not yet available): https://www.rcsb.org/structure/unreleased/8F76

The binding site does not consist of the residues identified by Man et al, meaning at least some ORs cannot be predicted using the current prediction attempts. All "method" and "corr" files located in the predict/ folder should now be assumed obsolete.

The binding of propionate in OR51E2 happens in a pocket delineated by F4.57 and L4.60 near the extracellular end of TMR4, aided by hydrogen bonds to Q45.53 and S6.55. H45.52 also appears to coordinate to the ligand, and is experimentally necessary for activation. The charged carboxylic acid group of propionate draws R6.59 nearer to TMR4, causing the extracellular end of TMR6 to rock inwards, accompanied by a counter motion of the cytoplasmic end. This appears to be what activates the receptor.

Propionate and acetate are small molecules that easily fit into this pocket. Larger carboxylic acids do not favor the active configuration. Mutagenesis of F4.57A causes the receptor to respond best to octanoic acid, while the L4.60A mutant prefers hexanoic acid, both in accord with the respectively larger pocket sizes. So it would seem that the ligand must make contact with a threshold surface area of the inside of the pocket, and be small enough to favor the inward motion of the top of TMR6, in order to be an agonist.

Accurate predictions will involve:
- Ensuring that the primarydock app correctly places the ligand into the pocket;
- Ensuring that agonist ligands favor the active state, defined by the rocking motion of TMR6;
- Testing this binding site, and other possible binding sites, for other empirical pairs;
- Identifying any other ways that a ligand might prefer the active position of TMR6.

Note that for thiol-sensitive receptors of the OR2 family, the presence of a copper binding site on TMR5 seems almost certain. How this will interact with the motion of TMR6 is another avenue to explore. Some ORs may have as yet undiscovered metal binding sites that may be revealed by a careful search.

This does not apply to TAARs, VN1Rs, or MS4As. TAARs probably function via their W6.48 trip switch, which ORs lack in favor of a FYG motif. The activation changes for VN1Rs and MS4As are still unknown.

ssepeq:
[...]

PR #214, opened 2023-03-17
Incremental rock. #200

ssepeq:
Mixed results with SCFAs in OR51E2. Predicts diacetyl correctly as non-agonist. Ionones and δ-decalactone are failing to make a complete dock.


ssepeq:
🟩🟩🟩

ssepeq:
Waiting on the server to process all of the empirical ligands for OR51E1 and 51E2. It will take a while.

From then, it will be a matter of modifying the method_ php to make predictions. Then it should be possible to get an accuracy estimate for carboxylic acids in OR51 receptors. Other ligands probably dock in a different part of the binding residues area; this may also be true of the tetrapod-like OR families.

All this calculation may ultimately require a rack mount server. They make ones now whose cooling fans don't sound like a jet engine. Tried looking into cloud computing, but anything in budget would be less powerful than the cheap used desktop I've been developing this project on.

primarydock.config
```diff
@@ -60,6 +60,18 @@ CEN RES 2.53 3.29 3.32 3.33 3.36 3.37 3.40 3.41 4.53 4.57 4.60 45.49 45.52 5.39
...
+# and/or transformations to be applied at specific nodes, without having to limit
+# them to a single active state node.
+# The below lines, if uncommented, will immediately move TMR6 2A in the +Y direction,
+# then rotate it 15 degrees for node 1.
+# Note all rotations and transformations are *cumulative*, not absolute; if given two
+# consecutive HXR lines each with a 5 degree rotation, the final rotation will be 10
+# degrees relative to the original protein structure.
+# HXR 0 TMR6 6.26 6.60 6.48 0 2 0 -0.55 0 0.84 0
+# HXR 1 TMR6 6.26 6.60 6.48 0 0 0 -0.55 0 0.84 15
+
```
primaryodors, line 74: May want to add something about axes based on a region, bw residue, or atom.

src/classes/constants.h
```diff
@@ -167,7 +167,7 @@
 #define drift_decay_rate 0.08
 
 // Allows full 360 degree whole molecule rotations to search for lower energy configurations.
-#define allow_mol_fullrot_iter 0
+#define allow_mol_fullrot_iter 1
```
primaryodors, line 170: Why is this enabled if axial tumble (line 157) is not?

src/classes/protein.cpp
```diff
@@ -800,40 +800,54 @@ void Protein::add_remark(std::string new_remark)
     // TODO: Sort remarks by number, becarefuling to preserve the sequence of same numbered remarks.
 }
 
-void Protein::set_clashables()
+void Protein::set_clashables(int resno, bool recursed)
 {
```
primaryodors, line 804: This looks like the same change as made in #213 - perhaps want to close the other PR?

test/51e2_propionate.config
```diff
@@ -17,13 +17,23 @@ LIG sdf/propionate.sdf
...
+PATH 3 REL 0 0 0
+PATH 4 REL 0 0 0
+PATH 5 REL 0 0 0
 
 SIZE 8.0 5.0 8.0
 
-ACVNODE 0
-ACVHXR TMR6 6.26 6.60 6.48 0 0 0 0.5015 0 0.8651 10?
+HXR 0 TMR6 6.26 6.60 6.48 0 2 0 bw 4.57 0
+HXR 1 TMR6 6.26 6.60 6.48 0 0 0 0.5015 0 0.8651 3
```
primaryodors, line 29: These lines should all match, either all BW or all Cartesian axis.

primaryodors (changes requested):

src/classes/constants.h
ssepeq: Will switch them in the morning and retest big three.

ssepeq: The TAAR test result is not as good if axial tumble is enabled and not full rotations.

ssepeq:
🟩🟩🟩

primaryodors (approved):

PR #215, opened 2023-03-28
Mandatory binding residues.

ssepeq:
Predicts good dock poses SCFAs up to 5 carbons for OR51E2, with longer chains correctly failing to dock. Predicts SCFA chains of 6, 8, 10, and 12 carbons for OR51E1 - of these, only C-12 seems to be a false positive. C-14 correctly fails to dock.

This is using the following config:

```
CEN RES 4.57 4.60 5.39
REQSR 4.57 4.60 5.39

SIZE 6.0 7.5 5.5

EXCL 1 56		# Head, TMR1, and CYT1.
WET
```


ssepeq:
image

Capric (decanoic) acid in OR51E1. Inactive state. The ligand (highlighted, red and gray) is secluded within the binding pocket, with its tail extending down towards the cytoplasmic end of the receptor (not shown). A negatively charged oxygen is exposed. If the top of TMR6 (orange helix) pivots towards roughly the direction of TMR2 (blue helix) and TMR3 (emerald green helix), then the Arg264(6.59) residue (highlighted, blue) can coordinate to the ligand and hold the receptor in the active state.

primaryodors:
Looks really good.

One thing that would be ideal would be to tie each REQSR record to a path node, and allow unlimited REQSRs.

ssepeq:
Testing the sweep6 method, capric acid rattles around way too much inside 51E1.

ssepeq:
There's a problem. OR51E2 is getting false positive predictions for acids C6 and C8-C10. Here's a screenshot that shows why:

image

Phe155 is one of the steric hindrance residues identified by electron microscopy and site directed mutagenesis that limits the size of acid molecules that can activate OR51E2. In the dock, it moves from its initial position (green arrow) to an incorrect configuration that clashes with atoms of TMR3 (red arrow). On examination of the ligand, if Phe155 didn't flex, there would be exactly enough space for an acid of up to C5 (valeric acid), which indeed is the longest straight chain SCFA identified as a 51E2 agonist.

Side chain flexions are happening despite causing internal clashes within the protein. In a case like this when a side chain cannot accommodate a feature of the ligand without increasing the protein's internal clashes, then probably the side chain should stay put and force the ligand to conform or result in a failed pose. In this case, a failed pose would be an expected outcome.

ssepeq:
The dock file in question:

[OR51E2-nonanoic_acid_sweep6.dock.txt](https://github.com/primaryodors/primarydock/files/11128469/OR51E2-nonanoic_acid_sweep6.dock.txt)

ssepeq:
From here: https://pubmed.ncbi.nlm.nih.gov/10737948/

> The number and identity of binding pocket residues that undergo side-chain conformational changes were determined. We show that, **in general, only a small number of residues in the pocket undergo such changes (e.g., approximately 85% of cases show changes in three residues or less**).

Emphasis mine.

This makes sense because structures downloaded from sites like AlphaFold or ZhangLab have already been optimized and determined to be the theoretical lowest total energy, or at least a local minimum near the lowest possible energy. Ligands are unlikely to influence any given side chain enough to change that optimal configuration. Exceptions exist, notably the Arg6.59 crucial for OR51E activation by SCFAs, so flex docking is still important. But the haphazard flexions currently produced by the docker are excessive and unrealistic.

primaryodors:
Once the problem of the ligand changing position is fixed, go ahead and do a Big Three test and mark this ready. The flexion probability fix will be on a separate branch.

ssepeq:
Ligand repositionings were occurring when a bond near the carboxylic acid end would rotate, instantly moving most of the carbon chain to a position almost 180deg away, followed by transformational motion of the entire ligand away from the newly clashing residues. That in and of itself is not correct behavior, and will have to be addressed eventually.

Disabling 360deg ligand bond rotations in all nodes except node 0 is showing promise. Testing with capric acid in OR51E1, the ligand stays roughly vertical throughout the "movie", forming an ionic contact with Arg264 in frame 110 (node 2, iteration 11) of the dock. From then, although the ligand does move around slightly, it maintains that ionic bond and does not do anything bizarre like suddenly turn sideways.

ssepeq:
🟥🟩🟩

Haven't identified which change breaks the amino aldehyde test, but it isn't turning off the 360deg switch.

ssepeq:
_momentum_rad_ceiling is the cause of the amino aldehyde fail.

primaryodors:
Should make _momentum_rad_ceiling a global variable and leave it high for the amino aldehyde test and limit it only for the best-binding dock.

ssepeq:
Still failing.

primaryodors:
The amino aldehyde test comes from a time when the best-binding method was not yet found to be the best way to position a ligand inside a binding pocket. The test is mostly superseded by ligand preparations specific to docking in a protein. That's why some of the protein docking features have to be turned off for the molecule test. Eventually, the amino aldehyde test might be replaced by something else, perhaps a test of coordinating a metal to a peptide.

If the TAAR8 and OR1A1 tests are still succeeding, feel free to merge this.

PR #216, opened 2023-03-31
--movie mode now allows multiple poses, nodes.

ssepeq:
The --movie option, useful for tracking docking progress by outputting each iteration as a discrete node to a temporary dock file, no longer stops cold after the zeroth node of the first pose. Now it will output every pose and, within each pose, all iterations of all path nodes sequentially. So for instance if your dock has 50 iterations and 3 path nodes, then it will output the iterations for path node 0 as movie noes 0 - 49, path node 1 as 50 - 99, etc.

Note that the pose numbers of the temp file are the sequence the poses were computed in, not the same as the pose numbers of the normal output, which are sorted by total energy.


ssepeq:
No change to docking algorithm.

PR #226, opened 2023-04-13
Another segfault fix.

ssepeq:
After calling hydrogenate() on a protein, some of the Molecule objects have an atcount that exceeds non-null Atom objects. This PR attempts a fix and implements a workaround.

No change to docking classes.


ssepeq:
🟥🟩🟩

ssepeq:
🟩🟩🟩

PR #230, opened 2023-04-16
Soft pocket indexed by pose, node.

ssepeq:
The previous functionality carried the same helix motions over all the poses and nodes, and applied the final motions to all output poses and nodes.

The new functionality separates them out by pose and node.


primaryodors:
Have to rename your branch. It should be descriptive of what the new code changes do.

PR #238, opened 2023-04-26
Progress bar

ssepeq:
Users can now specify `PROGRESS` in the .config file, or `--progress` in the command line, to see a progress bar on screen that indicates how far along the dock has come.

No change to the classes.


primaryodors:
Is there a way to erase the progress bar once docking has completed so it doesn't stay on the screen frozen at 99% despite the presence of dock results?

PR #239, opened 2023-04-26
Template homology

ssepeq:
Ability to use a known active state PDB as a homology guide to activate the protein for the dock.


ssepeq:
🟩🟩🟩

src/classes/atom.cpp
```diff
@@ -1262,6 +1247,11 @@ bool Bond::rotate(float theta, bool allow_backbone, bool skip_inverse_check)
         return false;
     }
 
+    if (atom->residue == 264 || btom->residue == 264)
```
primaryodors, line 1250: Let's remove this debug trap.

src/classes/constants.h
```diff
@@ -316,6 +316,8 @@
 #define _dbg_null_flexions 0
 #define _dbg_repack 0
 #define _dbg_multiflex 0
+#define _dbg_homology 1
```
primaryodors, line 319: Debug switches should be turned off before merging to stable.

src/classes/protein.cpp
```diff
@@ -2597,8 +2597,391 @@ bool Protein::disulfide_bond(int resno1, int resno2)
...
+
+            // "soft"-manipulate the region to minimize clashes.
+            Point ptx(dx[hxno], 0, 0);
+            move_piece(rgstart1, rgend1, (SCoord)ptx);
+            float after = get_internal_clashes(rgstart1, rgend1, repack);
+
+            if (after > before)
+            {
+                ptx.x *= -1;
+                // move_piece(rgstart1, rgend1, (SCoord)ptx);
```
primaryodors, line 2876: Dead code. There are two more instances of the same commented line below.

src/classes/protein.cpp
```diff
@@ -2597,8 +2597,391 @@ bool Protein::disulfide_bond(int resno1, int resno2)
...
+
+    // TODO: Should figure out how to do homology for the EXR and CYT loops. At minimum the EXR.
+}
+
+void Protein::bridge(int resno1, int resno2)
+{
+    AminoAcid *aa1 = get_residue(resno1), *aa2 = get_residue(resno2);
+    if (!aa1 || !aa2) return;
+
+    // MovabilityType mt1 = aa1->movability, mt2 = aa2->movability;
```
primaryodors, line 2962: Dead code.

src/primarydock.cpp
```diff
@@ -2236,6 +2305,48 @@ int main(int argc, char** argv)
...
+            return 0xbadf12e;
+        }
+        ptemplt->load_pdb(pf);
+        fclose(pf);
+
+        protein->homology_conform(ptemplt);
+
+        cout << "Adjusting helix positions...";
+
+        // TODO: This section should be moved to the Protein::homology_conform() ftn.
```
primaryodors, line 2325: Let's go ahead and do this code move before merging.

src/primarydock.cpp
```diff
@@ -2825,7 +2976,7 @@ int main(int argc, char** argv)
         delete protein;
         protein = new Protein(protfname);
 
-        if (hydrogenate_pdb)
+        if (hydrogenate_pdb || tplset)
```
primaryodors, line 2979: This should probably test that temp_pdb_file contains a string, or else use a flag inclusive of all features that generate a temp PDB.

src/primarydock.cpp
```diff
@@ -4243,6 +4406,8 @@ int main(int argc, char** argv)
                 if (debug) *debug << "Aligned ligand to AA." << endl;
                 cout << endl;
                 #endif
+
+                freeze_bridged_residues();
```
primaryodors, line 4410: There are 7 of these... are all of them really necessary?

src/primarydock.cpp
```diff
@@ -5078,7 +5244,7 @@ int main(int argc, char** argv)
         if (output)
         {
             hydrogenate_pdb = false;
-            pf = fopen(protfname, "r");
+            pf = fopen(tplset ? temp_pdb_file.c_str() : protfname, "r");
```
primaryodors, line 5247: Here too, should output any temp PDB if one has been created, not just homologized.

src/primarydock.cpp
```diff
@@ -5090,7 +5256,7 @@ int main(int argc, char** argv)
             fprintf(pf, "\nOriginal PDB:\n");
             protein->save_pdb(pf);
             fclose(pf);
-            cout << (hydrogenate_pdb ? "Hydrogenated " : "Original ") << "PDB appended to output file." << endl;
+            cout << (tplset ? "Homologized " : "Original ") << "PDB appended to output file." << endl;
```
primaryodors, line 5259: Can probably just drop the original/hydrogenated/homologized adjective.

test/big_three
```diff
@@ -10,6 +10,6 @@ clear; make code
 clear; bin/primarydock test/testTAAR8.config && play chime.mp3
 
 # Limonene test: Check for at least one successful binding pose with a total energy of less than zero.
-# *If using the ZhangLab PDB*, do a visual check using the viewer that Asp111 is not pulled awway from its salt bridge configuration with His243.
+# *If using the ZhangLab PDB*, do a visual check using the viewer that Asp111 is not pulled away from its salt bridge configuration with His243.
```
primaryodors, line 13: This line isn't relevant anyway since ZhangLab PDBs are no longer included with PrimaryDock.

primaryodors (changes requested):

src/primarydock.cpp
ssepeq: Removed a couple, but kept the ones that follow a load_pdb() call (definitely necessary) and things like best binding and activation matrix that change residue movabilities.

primaryodors (approved):

Issue #217, opened 2023-04-01
Side chain flexion probabilities.


primaryodors:
As mentioned in #215 , too many side chain flexions are happening during dock, including one flexion of a phenylalanine residue into a position where it clashes with nearby protein atoms, causing false positives in the docking of ligands too big for the binding pocket.

Here are the flexion probabilities from the data in the cited https://pubmed.ncbi.nlm.nih.gov/10737948/ :

[flexion probabilities.ods](https://github.com/primaryodors/primarydock/files/11129710/flexion.probabilities.ods)

As you can see, the probability of any given Phe residue flexing in any real world protein-ligand binding is 0.015. This means about one and a half percent of dock poses for random selections of protein and ligand should include any flexion whatsoever of any Phe residue.

The probability of any given Arg flexing is 0.0635. Since we know Arg6.59 always flexes in OR51E* activations by SCFAs, 100% of all OR51E* docks with agonist SCFAs should flex Arg6.59.

ssepeq:
If each pose be limited to a predetermined number of flexions, binding residues could be selected by their pre-iteration umet potential ligand binding times their flex probability. Then those residues, and only those residues, would be allowed to flex during the dock iterations.

The proposed "flexion selection" algorithm can begin with a 60% chance that at least one side chain will flex. If it does, then there would be a 60% chance that a second side chain would flex, and so on. This will produce a curve nearly identical to figure 2 of the article.

ssepeq:
After spot checking a sample dock file, there were 41 residues capable of interacting with the ligand, as indicated by the presence of PDB lines in the dock output. A binary flex/rigid on all binding residues equals 2^41 possible combinations, most of which have too many positive values.

For a dock with zero flexions, there is only one possible combination: all 41 side chains rigid.

For one flexion, there are 41 possibilities, which can be narrowed down by the algorithm.

For two flexions, there are 41*40 possibilities, or 1640. For three flexions, 41*40*39 = 63960. 15% of all dock poses will have even larger numbers of possibilities. Even though combinations can be narrowed down by flexion probability and unmet potential, the odds of getting the right combination by chance alone would be incredibly small.

It leads me to wonder how many poses will be required each dock in order to get at least one accurate result most of the time.

primaryodors:
So this feature got broken.

PR #244, opened 2023-05-06
Metal coordination in dock config file.

ssepeq:
A metal ion may now be specified in the config file to be added to the PDB before docking. It will act like an additional ligand, but with its coordinating residues frozen in place around it. The best-binding algorithm has been updated to treat a metal and its coordination residues as a single monolithic alignment target with a property of metallicity that favors sulfur atoms in ligands.

Also included are: increased flexibility of the ligand; reduced availability of backbone atoms for hydrogen bonding; an improvement to the viewer to avoid hiding side chains behind backbone cartoons.


ssepeq:
🟩🟩🟩

primaryodors:
Couple things with this PR...

The "sulfur kludge" is kind of antithetical to the design of PrimaryDock. Code shouldn't be tied to any specific element, just like we try not to tie code to specific amino acids (glycine being the exception since no other glycine-like α-amino acid is possible). Instead, we tie code to properties, such as the histidine flip which should work for any imidazole-containing AA etc. No element, other than perhaps hydrogen, is unique enough in all its properties to warrant coding for directly like this. How would a selenium-containing ligand behave in the real world under the conditions where this sulfur kludge would apply? It would be far better to specify "chalcogen within a specific electronegativity range" or even just "chalcogen" with electronegativity-specific feature multiplication factors rather than tying the code to a single element.

Similarly, commenting out the Cu...O coordination strength from bindings.dat is a poor way to deal with oxygen atoms coordinating with metal ions. Why wouldn't a ligand's hydroxy group coordinate to a Cu+ or a Cu2+? Could be the simulation is correct. If the journal article ran their simulation and found it does not coordinate, then there must be a documented reason why that can be used to inform PrimaryDock just as it was used to inform their docker. In that case, finding the answer and applying it to the code is more proper than willy nilly removing an interatomic binding potential simply because it was inconvenient.

Whatever reason would cause a hydroxy group to coordinate to other hydroxyls instead of to a metal, should also be the reason for not best-binding it to a metal "glom", cf. the sulfur kludge above.

As for the hydrogen bonding kludge, it can stay for now given the programming complexity involved in making H-bonding competitive. Going to start a new issue soon for adding that capability.

primaryodors:
This may be of help:

https://chem.libretexts.org/Courses/Saint_Marys_College_Notre_Dame_IN/CHEM_342%3A_Bio-inorganic_Chemistry/Readings/Metals_in_Biological_Systems_(Saint_Mary's_College)/Oxygen_transport_and_Storage/Blue_blood_(Invertebrates_molluscs_and_arthropods)/Haemocyanin

> Hard/Soft Acid/Base Theory (HSAB)
>
>Hard/Soft Acid/Base (HSAB) Theory allows for predictions of which acids and bases (metals and ligands) prefer to interact and which are most stable. In HSAB Theory, acids are electron acceptors and are referred to as acids/metal ions, and bases are electron donors and are referred to as ligands. These acids and bases can be categorized as hard or soft based on their charge density (charge to size ratio), polarizability (ability for an induced dipole to occur), and the covalent vs ionic nature of their interactions. Hard acids and bases have a higher charge density, are less polarizable, and/or have bonds that are more ionic; these tend to have a high charge and a small radius. On the other hand, soft acids and bases have a lower charge density, are more polarizable, and/or they have bonds that are more covalent; these tend to have a low charge and a large radius. HSAB can be used to predict which different metal-ligand pairs prefer to interact; hard acids prefer to coordinate with hard bases, soft acids prefer to coordinate with soft bases, and borderline acids prefer borderline bases. Essentially, the more “like” the characteristics of a metal and ligand, the stronger the preference of the ligand for that metal. This means that if there is a soft acid present,  Au+  for example, in the active site of a molecule, it will prefer to bind to a soft base, such as  S2− .10 In general, when you move down the periodic table, softness increases, and as you move right softness increases. Additionally, all of the bases are nonmetal electron donors (typically anions), and the acids are mostly metal ions.
>
> As seen in figure 2, the active site of hemocyanin contains two copper ions, each bound to three histidine (His) residues. When the Cu(I) ion binds to oxygen it is oxidized to Cu(II). Cu(I) is a soft acid, Cu(II) is a borderline acid, and the imidazole ring of histidine is a borderline base. Atoms or molecules that are classified as having intermediate characteristics are called borderline, as they do not lie at either extreme of the hard/soft spectrum; they are harder than metals that are categorized as “soft” and softer than metals categorized as “hard”. Nitrogens are typically classified as hard bases, however if a nitrogen is contained within an aromatic ring, the entire ring is more polarizable. For this reason, imidazole nitrogen atoms are classified as borderline bases. Because the copper ion alternates between a soft acid (Cu(I)) and a borderline acid (Cu(II)), it makes sense that it is bound to a borderline base. Borderline histidine is a better match for borderline Cu(II) than soft Cu(I), however His is still a better match for Cu(I) than a set of ligands that is harder. It is important to recognize, however, that there are other factors that affect metal ion selectivity, such as ionic size and ligand field stabilization energy, that must also be considered.

ssepeq:
https://doi.org/10.1002/anie.201007100

ssepeq:
This article in favor of HSAB gives numbers that would seem to predict that Cu2+ would preferentially coordinate to O rather than S: https://doi.org/10.1021%2Fja00364a005

primaryodors:
The question of HSAB vs. other methods warrants further study.

For this ticket, it is sufficient to code the following: Using the electronegativities of the ligand atom (El) and of the metal (Em), compute 0.5 + 0.5 * cos((El + Em) / 2 - 2.25) and use that as the multiplier for the ligand-metal coordination strength.

ssepeq:
The latest commit assumes that the metal ion is something with a higher electronegativity such as Cu. For olfaction, this might be a safe assumption, but for versatility, the best-binding algorithm should be rewritten to use potential binding energies between atoms instead of guessing with constant values for things like polarity, pi capability, charge, etc.

primaryodors:
Okay to go ahead with the potential binding energies method in this PR if it isn't too big a programming task.

ssepeq:
This is working currently:

```
PROT pdbs/OR2/OR2M3.upright.pdb
TEMPLATE pdbs/OR51/OR51E2.8f76.pdb
BRIDGE 6.57 45.51
LIG sdf/3-mercapto-2-methylpentan-1-ol.sdf

# https://doi.org/10.1007/s00018-019-03279-y
CEN RES 3.33 3.37 3.40 5.41 5.42 6.55
MCOORD Cu +1 5.42 5.43 5.46

SIZE 7.0 7.5 7.0
SOFT TMR2 TMR4 TMR5 TMR6 TMR7

EXCL 1 57		# Head, TMR1, and CYT1.
WET

POSE 10
ELIM 99
PROGRESS

FLEX 1
FLXR 3.33

ITERS 50

OUT output/%p_%l.active.dock
APPENDPROT
```

ssepeq:
This is the nonsense that it's pulling for glom pairing currently (using a thiolated ligand and after removing Met206(5.46) from metal coordination):

```
Final pair assignments:
atom_glom[ C1 C2 C3 C4 C5 C6 H9 H10 H11 H12 H13 H14 H15 H16 H17 H18 H19 H20 ]-residue_glom[ Cys202 Cys203 ]
atom_glom[ O7 H21 ]-residue_glom[ Cys202 Cys203 ]
atom_glom[ S8 ]-residue_glom[ Arg261 ]
```

ssepeq:
This PR is turning into another monster.

ssepeq:
TODO:

All the places where `multimol_conform()` is currently being called, should be replaced with calls to `conform_molecules()` instead. In cases where background molecules or molecules allowed to clash are specified, dedicated flags can be defined in `MovabilityType`.

Once all calls have been switched over, the monstrous `multimol_conform()` function and its myriad compiler constants can be removed.

All metal-binding code not pertaining to this PR should also be removed.

ssepeq:
Electronegativity changes with ionization charge. There's a formula in this paper: https://doi.org/10.1021/jp062886k that can be used to calculate an ion's electronegativity on the fly, so long as we add another dat file for information about ions, ionization energies, and crystalline ionic radii.

primaryodors:
Ion data would make a great topic for a future PR. Would like to get this PR wrapped up and ready to merge.

ssepeq:
🟩🟩🟩

primarydock.config
```diff
@@ -22,6 +22,19 @@ PROT pdbs/OR1/OR1A1.upright.pdb
...
 
+# Optional SMILES input string. This will override the ligand SDF file.
+# It is not recommended to specify a completely different molecule from the SDF file.
+# This feature is intended for instances where e.g. a ligand should be docked in an
+# ionized state and the SDF contains a neutral uncharged molecule, or an internal
+# arrangement of conjugated bonds in the desired ligand differs from the SDF, or any
+# other variation not involving the addition or removal of heavy atoms, nor isomerization
+# of the molecule.
+# Note that if the config file specifies a SMILES string, and a --lig argument is used
+# on the command line, that argument will *not* override the SMILES string from the config
```
primaryodors, line 33: Let's fix it so if the user does specify a --lig on the command line, it clears out the SMILES string.

src/classes/atom.cpp
```diff
@@ -1878,7 +1878,13 @@ float Atom::get_sum_pi_bonds()
     float retval=0;
     for (i=0; i<geometry; i++)
     {
-        if (bonded_to[i].btom && bonded_to[i].cardinality > 1 && bonded_to[i].cardinality <= 2.1) retval += bonded_to[i].cardinality;
+        // Aromatic ring fix breaks molecule test.
```
primaryodors, line 1881: Does the ring fix still break the two benzene molecules test?

src/classes/intera.cpp
```diff
@@ -699,7 +741,7 @@ float InteratomicForce::total_binding(Atom* a, Atom* b)
 
         if (!stacked_pi_rings)
         {
-            if (forces[i]->type == vdW
+            /*if (forces[i]->type == vdW
```
primaryodors, line 744: Dead code that can be removed.

src/classes/intera.cpp
```diff
@@ -864,14 +908,15 @@ float InteratomicForce::total_binding(Atom* a, Atom* b)
...
-                    if (theta > M_PI/2) continue;
+                    if (forces[i]->type != pi && forces[i]->type != polarpi) if (theta > M_PI/2) continue;
                     float contrib = pow(fmax(0,cos(theta)), dpb);
                     if (!isnan(contrib) && !isinf(contrib)) bsum += contrib;
                 }
 
                 if (asum > 1) asum = 1;
                 if (bsum > 1) bsum = 1;
 
+                // Removing these two lines fixes anisotropy but breaks the molecule test.
```
primaryodors, line 919: Is this still happening?

src/classes/molecule.cpp
```diff
@@ -1643,7 +1675,7 @@ void Molecule::crumple(float theta)
     int i;
     for (i=0; b[i]; i++)
     {
-        float ltheta = theta*randsgn();
+        float ltheta = frand(-theta, theta); // theta*randsgn();
```
primaryodors, line 1678: Dead code at end of line.

src/primarydock.cpp
```diff
@@ -3620,31 +3561,47 @@ int main(int argc, char** argv)
...
                                     )
+                                    {
                                         too_similar = true;
-                                    // if (sc_gloms[j].get_center().get_3d_distance(glomtmp.get_center()) < 0.01) too_similar = true;
-                                    float rlg = ligand_gloms[j].get_center().get_3d_distance(ligand_gloms[l].get_center());
+                                        #if _dbg_glomsel
+                                        if (too_similar) cout << "Residue belongs to an existing glom." << endl << endl;
+                                        #endif
+                                    }
+                                    // TODO: WTF was this for??????
```
primaryodors, line 3599: Language.

src/primarydock.cpp
```diff
@@ -3620,31 +3561,47 @@ int main(int argc, char** argv)
...
+                                        if (too_similar) cout << "Potential glom contains an already assigned residue." << endl << endl;
+                                        #endif
+                                    }
                                     if (glomtmp.aminos.size() == 1
                                         &&
                                         std::find(sc_gloms[j].aminos.begin(), sc_gloms[j].aminos.end(), glomtmp.aminos[0]) != sc_gloms[j].aminos.end()
                                     )
+                                    {
                                         too_similar = true;
-                                    // if (sc_gloms[j].get_center().get_3d_distance(glomtmp.get_center()) < 0.01) too_similar = true;
```
primaryodors, line 3638: Dead code.

src/primarydock.cpp
```diff
@@ -4124,7 +4093,7 @@ int main(int argc, char** argv)
                                 mtmp[2] = nullptr;
                                 m.movability = MOV_FLEXONLY;
                                 alignment_aa[l]->movability = MOV_FLEXONLY;
-                                Molecule::multimol_conform(mtmp);
+                                Molecule::conform_molecules(mtmp);
                                 m.movability = MOV_ALL;
                                 // m.intermol_conform_norecen(alignment_aa[l], iters, reaches_spheroid[nodeno]);
```
primaryodors, line 4098: Dead code.

src/primarydock.cpp
```diff
@@ -4242,16 +4218,33 @@ int main(int argc, char** argv)
...
                 iters,
                 &iteration_callback
-            );
-            delete[] delete_me;
+            );*/
+            ligand->movability = (MovabilityType)(MOV_ALL - MOV_MC_AXIAL);
+            Molecule::conform_molecules(cfmols, iters, &iteration_callback);
+            // delete[] delete_me;
 
             /*time_t jlgsux = time(NULL);
```
primaryodors, line 4249: Should remove this.

src/primarydock.cpp
```diff
@@ -4643,6 +4648,17 @@ int main(int argc, char** argv)
...
+                {
+                    mtlcoords[l].mtl->stream_pdb_line(
+                        pdbdat,
+                        9900+l
+                    );
+                }
+            }
+
             dr[drcount][nodeno].pdbdat = pdbdat.str();
             // cout << "Attempt " << drcount << " node " << nodeno << " pdbdat is " << dr[drcount][nodeno].pdbdat.length() << " chars." << endl;
```
primaryodors, line 4663: Dead code.

viewer.htm
```diff
@@ -1749,6 +1753,7 @@
     if (curr_ligand) $('#ligbtn, #prsv').show();
     
     // if (cen_ligand) curr_ligand.autoView();
```
primaryodors, line 1755: Dead code.

viewer.htm
```diff
@@ -1828,6 +1833,7 @@
...
```
primaryodors, line 1835: Dead code.

viewer.htm
```diff
@@ -1869,6 +1875,7 @@
 
 	// cc.setSelection("*");
```
primaryodors, line 1876: Dead code.

viewer.htm
```diff
@@ -2298,7 +2305,7 @@
 						stage.animationControls.rotate(quaternion);*/
```
primaryodors, line 2305: Dead code.

primaryodors (changes requested):
Let's also make sure the MCOORD command in Pepteditor works the same as the MCOORD parameter in PrimaryDock, i.e. they should both use the new metal coordination code.

Later, we'll be using a command line utility to identify and remove obsolete functions, and the above step will be helpful for that.

ssepeq:
> Let's also make sure the MCOORD command in Pepteditor works the same as the MCOORD parameter in PrimaryDock, i.e. they should both use the new metal coordination code.
> 
> Later, we'll be using a command line utility to identify and remove obsolete functions, and the above step will be helpful for that.

They're different formats with different capabilities. The amount of development involved would deserve its own PR. Recommend adding an obsolescence notice to the Pepteditor MCOORD and ultimately replacing it with some kind of all-purpose atom/molecule coordination/prosthetic group feature, i.e. something that could add not only metal ions but also retinal, heme, etc.

primaryodors:
The deprecated notice will be fine.

src/aniso_test.cpp
```diff
@@ -93,13 +92,16 @@ int main(int argc, char** argv)
     }
 
     InteratomicForce* hb=0;
+    float strongest = 0;
     for (x=0; ifs[x]; x++)
     {
         cout << ifs[x]->get_type() << endl;
-        if (ifs[x]->get_type() == hbond)
+        //if (ifs[x]->get_type() == hbond)
```
primaryodors, line 99: Dead code.

src/aniso_test.cpp
```diff
@@ -93,13 +92,16 @@ int main(int argc, char** argv)
...
+        if (ifs[x]->get_kJmol() > strongest)
         {
             hb = ifs[x];
-            break;
+            strongest = ifs[x]->get_kJmol();
+            // break;
```
primaryodors, line 104: Dead code.

src/aniso_test.cpp
```diff
@@ -173,7 +175,8 @@ int main(int argc, char** argv)
                 oxy.clear_geometry_cache();
 
                 float tb = InteratomicForce::total_binding(anisoa, &probe);
-                tb /= hb->get_kJmol();	// This is not working.
+                tb /= hb->get_kJmol() * 1.5;
+                // cout << tb;
```
primaryodors, line 179: Dead code.

src/aniso_test.cpp
```diff
@@ -190,7 +193,7 @@ int main(int argc, char** argv)
                     clear_color();
                 }
 
-                if (!x && !y)
+                /*if (!x && !y)
```
primaryodors, line 196: What was the reason for commenting this block?

src/primarydock.cpp
```diff
@@ -1227,19 +1004,40 @@ int interpret_config_line(char** words)
...
+        i++; if (!words[i]) throw 0xbad372;
+        mcr.charge = atoi(words[i]);
+
+        i++; if (!words[i]) throw 0xbad372;
+        for (; words[i]; i++)
         {
             if (words[i][0] == '-' && words[i][1] == '-') break;
-            mcoord_resno[j++] = atoi(words[i]);
-            optsecho += to_string(atoi(words[i])) + (std::string)" ";
+            // mcoord_resno[j++] = atoi(words[i]);
```
primaryodors, line 1032: Dead code.

src/primarydock.cpp
```diff
@@ -1227,19 +1004,40 @@ int interpret_config_line(char** words)
...
+            // optsecho += to_string(atoi(words[i])) + (std::string)" ";
+            ResiduePlaceholder rp;
+            rp.set(words[i]);
+            mcr.coordres.push_back(rp);
         }
-        mcoord_resno[j] = 0;
+        // mcoord_resno[j] = 0;
```
primaryodors, line 1038: Dead code.

primaryodors (changes requested):
Just a few more items.

src/aniso_test.cpp
ssepeq: Because that code was causing this garbage to happen:

image

primaryodors: Do you see why it's doing that?

ssepeq: Yes, and I fixed an unrelated geometry bug.

primaryodors (approved):

primaryodors:
🟩🟩🟩

ssepeq:
Drat. The side chain flexion stuff from #217 seems to have come undone with this PR. Even metal coordinating residues are flexing willy nilly.

primaryodors:
This is the kind of thing that should be tested before marking the PR ready for review!

PR #247, opened 2023-05-17
Keep separate heavy and non-heavy rot bond caches.

primaryodors:


primaryodors:
Testing new updated cache feature.

Then must test Big Three and, once everything is fixed, merge.

primaryodors:
🟩🟩🟩

However, the metal coordination residues are still flexing.

I will continue this tomorrow.

Issue #248, opened 2023-05-18
Rename "glom" to Group.


primaryodors:
Even if "glom" in this context means conglomeration, it would be more professional to have the classes and variables - and documentation - refer to these groupings as groupings or simply as groups.

Giving this high priority since it should be a quick easy task.

PR #253, opened 2023-05-21
Rename glom to group. #248

ssepeq:


ssepeq:
🟩🟩🟩

src/classes/aminoacid.cpp
```diff
@@ -926,8 +926,8 @@ Point* AminoAcid::predict_next_NHCA()
     return retval;
 }
 
-#define _COUT_GLOMANOM 0
-void AminoAcid::glom(Point* predicted, bool CO)
+#define _dbg_conglomerate 0
+void AminoAcid::conglomerate(Point* predicted, bool CO)
```
primaryodors, line 930: What does this do?

ssepeq: Moves the amino acid so that its N terminus or C terminus line up with the neighboring amino acid in the sequence.

It should probably be called attach or something similar.

primaryodors: Let's rename this function to `attach`.

primaryodors (changes requested):

primaryodors (approved):

Issue #255, opened 2023-05-24
Reduce number of residues in dock output.


primaryodors:
Currently, we output all the residues that were used for the conformational space search, along with any that have flexed. This produces a crowded view of the environment around the ligand.

Let's exclude from the output any residues, that have not flexed, and are not bridged, that have no atoms within 5Å of any ligand atom.

PR #261, opened 2023-05-29
Limit output residues to 5A and flexed.

ssepeq:
Code changes for #255 .


ssepeq:
Docking itself is not affected since the only changes to the classes are 1.) adding a function to `Molecule` to enable the output portion of the primarydock app to measure intermolecular distances after the fact, and 2.) bringing back the functionality of the `been_flexed` flag, which was lost with the switch from the old multimol function to `conform_molecules()`.

primaryodors (approved):

Issue #257, opened 2023-05-28
Create Master and Express unit tests.


primaryodors:
>> It would appear that the Big Three on their own are not a sufficient test for each PR before merge. The ideal would be to have a master unit test that can be left running (overnight, if necessary) before merging, and a subset unit test of the most easily breakable things, that can be run periodically during development.

_Originally posted by @primaryodors in https://github.com/primaryodors/primarydock/issues/254#issuecomment-1566171331_

Let's create some kind of Master unit test that not only checks the Big Three, but also checks:
- Valid anisotropy results for several interacting atom pairs;
- Strengths of various noncovalent bonds;
- Dimerism of other molecule pairs besides ethanol-ethanol;
- At least one active soft dock, including Y6.55 coordination and R/K6.xx-D/E45.51 salt bridging;
- Atomic similarity;
- Amino acid similarity;
- Atom group selection;
- Amino group selection and group pairing;
- Amino acid tests;
- Point object tests;
- Protein test.

The Master unit test will be required for all PRs before merge, unless the PR does not affect any of the calculations in the classes, apps, or tests. E.g. a simple UI change would not require the Master test.

Let's also create an Express unit test containing only the most easily broken features, designed to not take more than a minute or two. (Details TBD.)

Importantly, the new unit tests would not stop once an error is reached, but would instead test each item in sequence and report a pass/fail for each item, then error out only at the end if one or more items failed.

PRs marked ready for review before the new tests are merged in, will not be held up by the new requirements. In other words, what's currently in review can proceed to approval and merge without having to wait for the unit tests to be finished.

ssepeq:
We can add metal coordination to the test. #254 broke metallic interactions; #261 has a fix.

primaryodors:
Let's resume these unit tests next; having them written and working will make it immensely easier to maintain the codebase and minimize bugs.

PR #263, opened 2023-05-29
Unit tests.

ssepeq:
Created a master unit test that runs the big 3 tests. TODO: Check output of docks.

Plenty more work to be done on this. See #257 .


primaryodors:
Any new progress on this? @ssepeq

ssepeq:
Going to be afk for several days this week and next. Want to get protein coupling working reliably and then have coupled PDBs ready and docking running on a cron on the local server, that way there'll be docks ready for analysis on my return.

primaryodors:
Going to want to include `test/motif_test.pepd` in the master unit test.

primaryodors:
While the server is performing large numbers of docks, now would be a great time to resume the unit tests.

ssepeq:
Molecular assembly is a crazy bizarre funhouse right now! Here's one example:

image

Another bug is saturated rings (cyclohexane etc) aren't working. It makes a straight chain hydrocarbon and then bonds the ends together, but doesn't curl it into a ring. I think that one might be a not too difficult fix in order to get mol_assem_test working.

primaryodors:
Don't worry too much right now about the molecule assembly tests that are failing; recall that some SMILES strings have never worked in the classes and that's why we integrated with OpenBabel.

Any tests that are failing should be left commented out for now. If a test can be fixed with just one or two lines of code, that's fine to include, but anything more than that is out of scope for this PR.

ssepeq:
☹️ 

Turning off debugging and checking in molecule assembly changes so far...

primaryodors:
These tests are looking good. If you can think of anything else to add, go for it, otherwise this master test looks good enough to start with.

The express test can include everything except the TAAR8 and OR1A1 tests. Once that is ready, feel free to mark this for review.

ssepeq:
Even the protein test and motif test? They each take a moment to run.

primaryodors:
The protein test, motif test, and molecule assembly tests can be omitted from the express test.

primaryodors (approved):

primaryodors:
The Big Three test is now obsolete; all future PRs will be required to pass unit_test_master (i.e. it should not output any red text) before merge.

PR #283, opened 2023-06-24
Prioritize certain pairs in cron for test purposes.

ssepeq:
It's going to take too long waiting for the cron to get through OR1A1, then OR1A2, and so on in sequence. Here are some sample test pairs representing receptor ligand diversity, as well as details and nuances like similar molecules that have different activities for a receptor and similar receptors that have different activities for a ligand.

Hopefully this will give a clearer estimate of prediction accuracy without having to wait for hundreds of agonists.


primaryodors:
Good idea!

PR #286, opened 2023-06-25
More cron stuff.

ssepeq:
The cron got stuck on OR3A1 making zero progress so these changes attempt to remedy that.

PR #289, opened 2023-07-04
couple.pepd

ssepeq:
Based on the same branch as #286 .

Creates a new file called couple.pepd to handle G-protein coupling in Pepteditor. Adds the capability for the editor to handle multiple protein strands.

Note this breaks compatibility with earlier versions because the magic variables `$SEQUENCE`, `$SEQLEN`, and the region start/end and BW50 variables now all include the strand ID in the name.


primaryodors:
What's the status of this and the other PR? @ssepeq

PR #303, opened 2023-07-15
Prediction code from failed branch.

primaryodors:


primaryodors:
Few things this should be doing better:

1.) In the hGNAQ coupling, Asn356 near the C terminus of the G protein is clashing with Thr295 of the 8th helix of the receptor. When transforming or rotating TMR7, the motion should include the entire tail region of the receptor, except in a very few cases of receptors that have an 8th transmembrane region.

image

2.) Asn356 has a 3HD2 atom, which is incorrect. The nitrogen atom of an amide should have only 2 hydrogens.

3.) Also in the hGNAQ coupling, TMR6 is being moved too far away from TMR7 at the cytoplasmic end. 

image

primaryodors:
This is making predictions, though not good ones (yet).

PR #323, opened 2023-08-07
Activation residues.

primaryodors:
Based on the work of Zhou et al, 2019 (https://doi.org/10.7554/eLife.50279); and Trzaskowski et al, 2012 (https://doi.org/10.2174/092986712799320556); and the cryo-EM models of mTAAR9 (https://doi.org/10.1038/s41586-023-06106-4) and hOR51E2 (http://doi.org/10.2210/pdb8f76/pdb), this PR aims to find sidechain and backbone activation changes applicable to each OR and TAAR, in order that the activation conformation changes can be predicted.


primaryodors:
One cause for concern is if phenethylamine is docked in the active and inactive mTAAR9 models, without flexion, the inactive binding energy is consistently more favorable than the active.

primaryodors:
Class I ORs nearly all have S3.40 and Y6.48, which form a slightly stronger hydrogen bond during activation. This is likely to be the basis of TMR6's pivot motion about 6.48. Some class II ORs also have S3.40 and Y6.48, including OR1A1 which must function by a different motion of TMR6 since an inward pivot of TMR6's extracellular end would result in a worse affinity for the strong agonist d-limonene. ORs which lack a hydrogen bonding 3.40 and/or 6.48, such as OR1G1, probably have a different pivot point for TMR6, perhaps the extracellular end of the helix.

GPCRs seem to be of three types re TMR6 motion: pivot at 6.48 like OR51E2, pivot at the EXR end of TMR6, and outward curling of TMR6 as in mTAAR9. OR1A1 might be of the third type.

Nearly all GPCRs have Y5x58 forming a hydrogen bond with Y7x53 in the active state, associated with the inward bend of TMR7 toward TMR3. The vast majority of ORs also have H6.40, which maintains an h-bond with Y5.58 in the active and inactive states.

The side chain of I5.62 moves to a place between 6.34 and 6.37 during activation, with repacking of 6.37. Together, these three residues probably determine the exact positioning of the CYT end of TMR5 in the active state.

L3.43 occurs in the majority of ORs. It seems to function to limit the inward motion of TMR7 in the cytoplasmic domain. Some OR10 receptors, including all or nearly all of OR10G, have Y3.43, which makes an h-bond with N7.49 in the inactive state. It is not known how this affects activation. The macrocyclic musk receptor OR5AN1 has M3.43.

These are some of the residues that may help determine the characteristic motions of each receptor's active state. ORs with differing side chains in these positions probably have differences in the shape of the active configuration.

primaryodors:
TAAR activation depends on either forming or breaking a contact between Y7.43 and D3.32. In class I ORs, the contact seems to be between 7.42 and 3.32, both of which are usually hydrophobic. Class II ORs have more variation in these two locations, and may or may not form a contact. In OR51E2, this seems to define the pivot point for TMR7, with the EXR end bending outward and toward TMR6 and the CYT end bending slightly toward TMR3.

6.51 is always F or Y in TAARs, but is not conserved in ORs. Its function remains to be elucidated here.

TAARs maintain a salt bridge between 7.36 and 2.64, however ORs have no such contact, and in fact the EXR regions of TMR7 and TMR2 actually move apart somewhat in active OR51E2. Compatibility and reachability of these side chains may therefore determine the presence or absence of a TMR7 EXR bend.

Both mTAAR9 and OR51E2 show a decreased contact strength between N7.49 and D2.50 in the active state. Since these are both highly conserved residues, variations might result in a nonfunctional receptor.

primaryodors:
In the inactive state, T6.36 of the KAFSTCxSHL motif and R3.50 of the MAYDRYVAIC motif form a hydrogen bond. This is disrupted in the active state. Receptors that lack this bond, such as OR1G1, may have enhanced responses to odorants since there is less energetic stabilization of the inactive configuration.

primaryodors:
OR1G1 has few internal contacts in the inactive state. N7.57 and Y2.40 form a hydrogen bond; these residues are well conserved in class II ORs, whereas class I ORs usually have T7.57 with no hydrogen bond to 2.40.

D1.60 forms an ionic contact with K303 on helix 8. Both residues are generally conserved.

M3.46 and M2.39 form a hydrophobic contact. Both residues are part of conserved motifs (PMYFFL and MAYDRY).

V7.43 and V2.54 form a weak hydrophobic contact.

F6.48 contacts V5.51 and L5.54, while F6.47 is nestled in between V3.36, E3.39, A3.40, and L3.43. F6.47 seems like the most logical pivot point, and F6.48's side chain might rotate downward during activation to fit between L5.54 and L3.43. The residue in position 6.55 is a D, which could make contact with N45.53 in the active state; the two residues are placed very far apart in the inactive conformation, meaning TMR6 has plenty of room to pivot. The CYT3 loop is short, so the CYT end of TMR5 would follow that of TMR6 closely. There's nothing in the way to prevent a wholesale pivot of TMR5 from its EXR end, or perhaps from I5.39 which contacts I45.52 and L45.56.

The relative lack of internal contacts in OR1G1 probably makes for an exceptionally flexible binding pocket that can accommodate a variety of ligand shapes and would have to be "soft" docked for purposes of predictions. This combined with the relative freedom of TMR5 and TMR6 to move almost unencumbered could be the reason why this receptor has a large agonist repertoire.

primaryodors:
The code changes in this PR fix a number of minor issues and should be merged as soon as possible.

The findings detailed here can be used to write code that, when given a receptor ID, analyzes the receptor's sequence and 3D structure to determine which activation features are present. That code will then allow the automatic generation of a set of DYNAMIC parameters for optimized docking.

PR #326, opened 2023-08-13
Additional improvements for TAAR8 cadaverine test.

primaryodors:
The AtomGroup::get_potential_ligand_groups() function now imposes a limit on its output that no two groups may share the same atom. For the TAAR test, this results in the two amino moieties each getting their own AtomGroup, with the aliphatic chain residing in a separate AtomGroup. The result of the improved group assignments is that the best binding algorithm has a strong tendency to assign the aliphatic chain to Cys115, placing it in the right position to force W6.48's side chain into a lowered position near F5.47, likely engaging the receptor activation mechanism.


primaryodors:
Running unit tests.

PR #327, opened 2023-08-14
STCR parameter fix.

primaryodors:
Previous behavior was when a .config file included a STCR parameter, the PrimaryDock app did not always obey the no flexion rule for the specified residue(s).

This PR fixes that behavior so that residues immobilized with a STCR param will remain unflexed.


primaryodors:
Unit tests passed.

PR #330, opened 2023-08-19
Active Dynamic Predictions

primaryodors:
Predicts active conformations using dynamic region motion parameters generated from the receptor's sequence, and then uses the predicted active states to make ligand activity predictions.

TODO:
- Whole-helix translations based on contact between residue side chains (e.g. OR1A1:Phe206-TMR6 clash avoidance);
- Once the above step is complete, activation motions should be cached for each receptor;
- Predictions should be run for several class I and class II ORs and some of the TAARs (waiting on a new server for this);
- PREDICTIONS.md to be updated with new method_optimized.


primaryodors:
All tests successful.

primaryodors:
Dynamic "optimization" takes vastly too much processing time and only converges back to the inactive protein state. We're taking this in a slightly different direction, where instead of attempting to optimize a set of dynamic motions, the motions will be applied once and saved to a temporary or cached PDB. It will be called Contact Residue-based Active-state Modeling.

A new utility will be created named ic_active_pdb and it will receive the output from the icactive.php script, apply all the DYNAMICs at1.0 effect to the input PDB, and save the output PDB as `pdbs/{family}/{receptor}.icactive.pdb`. Then a new `method_icactive.php` will dock the specified ligand into the specified receptor's .upright.pdb and .icactive.pdb models, deriving its prediction from the result.

PR #331, opened 2023-08-19
Aldehyde flip.

primaryodors:
Aromatic aldehydes can freely rotate their CHO group, but only by 180 degrees.

It is hoped that this fix will correct the prediction for ethyl vanillin's inverse agonism of OR1A1, by facilitating a hydrogen bond to Tyr178.


primaryodors:
All tests successful.

PR #332, opened 2023-08-20
Contact residue based active state modeling.

primaryodors:
Based on the now obsolete #330 .

TODO:
- Whole-helix translations based on contact between residue side chains (e.g. OR1A1:Phe206-TMR6 clash avoidance);
- Change method_optimized to method_icactive;
- R6.59 should receive an ATOMTO so that it coordinates to the ligand;
- Predictions should be run for several class I and class II ORs and some of the TAARs (waiting on a new server for this);
- PREDICTIONS.md to be updated with new method_icactive.


primaryodors:
This PR does not contain a facility for generating active conformations of TAARs; writing one will involve examining the mTAAR9 models in detail and the existing notes to come up with a sequence of changes like we have in `ic_activate_or.cpp`.

There is a server set up and running, but it is not very fast and it will take a while to process a sample of receptor-ligand pairs. Preliminary results are showing very low prediction accuracy, but with the potential for much higher accuracy if the relative positioning of ligands in receptor binding pockets can be slightly improved.

PR #335, opened 2023-08-25
OR51E2 Predictions.

primaryodors:
This PR restores the earlier functionality from when predictions for OR51E2 for short chain fatty acids (SCFAs) were approaching 100% accuracy.

Testing is underway on the remote server. Only butyric and isovaleric acids have generated incorrect results.

All unit tests have passed.


primaryodors:
The horizontal motion of TMR5 is ignoring a clash between D5.50 and S3.40. In the cryo-EM model, S3.40 prevents Y5.58 from moving all the way to reach Y7.53, but in the predicted model, the TMR5-TMR7 contact is made, moving TMR5 too much, closing the binding pocket too tightly around the ligand, and producing a false negative result for butyric, valeric, and isovaleric acids.

primaryodors:
All tests succeeding.

PR #337, opened 2023-08-29
Predictions on web app, phase 1.

primaryodors:
For receptors that have predictions, a new column appears on the receptor page. It gives a relative prediction score indicating how strong an agonist (positive value) or inverse agonist/antagonist (negative value) the docker thinks each ligand is for the receptor. The value is linked to a page that shows the active-state prediction model (if predicted to be a positive agonist) or the inactive-state dock (if predicted to be a negative agonist) in the 3D viewer.

Since predictions are still exhibiting low accuracy, a note appears on the page that predictions are in beta, and that the user is invited to contribute pull requests to the repository.


primaryodors:
No change to docking code.

PR #340, opened 2023-09-01
Receptor specific pockets.

primaryodors:
Uses a JSON file to manually designate ligand binding pockets for receptors individually.


primaryodors:
All tests passing.

PR #349, opened 2023-09-08
Download links.

primaryodors:
A download icon appears next to the prediction, with a popup menu to allow downloading the active and inactive models and docks, and a JSON file of the result.


primaryodors:
No changes to docking code.

PR #357, opened 2023-09-17
Separate active and inactive dock config files.

primaryodors:
Write separate config filenames to tmp/ for active and inactive docks.

Also removes obsolete prediction methods, adds a quick and easy shell script for command line predictions, and adds a striped appearance to the tables on the www app.


primaryodors:
Of course the fix didn't solve the problem.

PR #359, opened 2023-09-18
Flexion enhancements.

primaryodors:
Increases the propensity for bonds to rotate to optimize non-covalent bonding, especially when a bond rotation moves only hydrogen atoms.


primaryodors:
Testing a prediction before running unit tests. If the dock results look good, i.e. no missed connections due to atoms facing the wrong way, then this PR is successful even if the prediction is wrong. If there are still missed connections, then the PR is not finished even if the prediction is correct.

primaryodors:
54 minutes and only 70% done.

primaryodors:
All tests passing except the hexenol prediction, which is obsolete.

PR #362, opened 2023-09-25
PDB scoring tool.

primaryodors:
This PR adds a utility `score_pdb` which accepts an input PDB containing a ligand, scores the interaction energy between the protein and ligand, and outputs the result in the same format as the `primarydock` app.

It is useful for examining the ligand binding energies of experimental X-ray or cryo-EM structures, as well as debugging models generated by the docker that contain undetected clashes.


primaryodors:
Removed the OR1A1 hexenol unit test. It fails too often to be a useful indicator of a PR fit to be merged.

PR #366, opened 2023-09-27
Improved extracellular Y6.55~D/E45.51 contact for Hybrid6 activation.

primaryodors:
Receptors that have both Y6.55 and D/E45.51, that do not have a hydrogen bond between these two residues in the inactive configuration, are currently predicted to be Hybrid6 receptors meaning TMR6 performs some amount of full-helix pivot about 6.48 and 3.40 in addition to a Gly6.49 bend. This PR enhances the receptor activation utility to make a strong contact more often between 6.55 and 45.51 in receptors that were not successfully making this hydrogen bond before.

Should test this on more receptors, including OR2AT4. It seems to work well for OR10G7, making the contact and then correctly predicting agonism for eugenol and non-agonism for decanal (technically should be inverse agonism, and a small adjustment to the prediction method would fix this). Nonanal should be giving a correct prediction of non-agonist, except the docking code is incorrectly allowing the ligand to self-clash in the active model and wrongly causing a prediction of agonist. The self clash and prediction algorithm adjustment are beyond the scope of this PR.

Also beyond the scope of this PR is many receptors, including OR10G7, continue to fail to coordinate Y5.58 with Y7.53 via hydrogen bonds to an intermediate water molecule. Some receptors may display worse 5.58-7.53 distance with this PR, however since the functionality isn't working yet, this PR doesn't technically break it.


primaryodors:
Running master unit tests because these code changes might affect the OR51E2-propionate prediction.

primaryodors:
Unit tests succeeded.

primaryodors:
OR2AT4 has Y6.59 that would probably contact D45.51. 6.55 is also Y but is not positioned correctly to make the usual contact.

OR2AT4 is predicted to be a Rock6 receptor, so it is beyond the scope of this PR.

PR #367, opened 2023-09-28
Clash limits.

primaryodors:
- Outputs ligand self clash as part of each dock result.
- Filters dock results so that the ligand internal clash is never more than individual_clash_limit defined in constants.h.
- Filters dock results so that the least favorable residue-ligand binding energy is not more than individual_clash_limit plus the total energy limit from the config file.


primaryodors:
All tests passing, though propionic had to be run twice.

PR #369, opened 2023-09-30
More output model fixes.

primaryodors:
Tracking down a very strange bug.

The `tmp_pdb_residue` array, a stack allocation, is populated with the positions of residue atoms (using .copy_state()) very soon after docking a pose. Later on, the residue atom positions may have changed (TODO: why?) so their positions are returned to those from the dock result using .restore_state().

To prevent any possibility of mangling the locations of atoms, the molecule passed to ::restore_state() must be the same molecule or a compatible molecule to that passed to ::copy_state().

The problem is if a dock is running for one protein, say OR2F1, and other docks are running for a different protein, such as OR51E2, the `tmp_pdb_residue` array for OR2F1 is getting corrupted with residues from OR51E2 - from an entirely separate process - and ::restore_state() is failing because it "thinks" the atoms should be OR51E2's atoms.

### DIFF
--- a/data/binding_pocket.json
+++ b/data/binding_pocket.json
@@ -1,4 +1,14 @@
 {
+    "OR2F*":
+    {
+        "odorophores":
+        {
+            "OC(C)C":
+            {
+                "pocket": "3.33 4.60 5.46"
+            }
+        }
+    },
     "OR2M3":
     {
         "odorophores":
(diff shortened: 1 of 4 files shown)


primaryodors:
Any time an input PDB file is modified by the `primarydock` app, such as hydrogenation or the addition of a metal ion, a temporary PDB file is created. Both OR2F1 and OR51E2 are bridging residues together, triggering this temporary file which was being overwritten with one or the other receptor.

The temp file is now unique to each process, containing the pid in the filename. The problem does not seem to be reoccurring.

primaryodors:
No changes to docking code.

PR #373, opened 2023-10-08
TMR5-TMR6 extracellular end clash fix.

primaryodors:
Fixes a clash where some receptors, e.g. OR52D1 upon activation with `bin/ic_activate_or`, show a clash between residues of TMR5 and TMR6 near the extracellular domain.

PR also contains fixes to clash detection and a rename of a constant whose scope of use now exceeds its former name.


primaryodors:
Amino test is failing.

Also there is a step in `ic_activate_or` that moves TMR5 sideways toward TMR6. If the resulting model has TM5-6 clashes, then this step is not doing enough to ensure it stays within its limit. The cytoplasmic TM5-TM7 tyrosine coordination (which by the way involves a water molecule that still is not being modeled) can be achieved not only by translating the entire TM5 region but also by pivoting it about its sterically limiting contact like TMR6 does for OR51E2. Pivoting TMR5 might open the binding site to known agonist ligands that do not fit the inactive model.

primaryodors:
OR51E2-propionate test failed.

primaryodors:
Should come up with a better amino test.

All other tests succeeding.

primaryodors:
TAAR test keeps failing; all other tests succeeding.

primaryodors:
TAAR test is failing in best binding. Only the amine-Asp111 pair is being used; the amine-Asp201 pair is ignored.

PR #377, opened 2023-10-18
Leave histidines uncharged and nonpolar unless H-bonded to an acid.

primaryodors:
Histidine side chains will now default to their non-protonated form, which acts like a hydrophobic aromatic residue (e.g. tyrosine or phenylalanine). Only if the side chain can form a hydrogen bond with a negatively charged atom (with or without performing a "histidine flip" where the movable hydrogen relocates to the δ nitrogen position) can the histidine become protonated and polar.


primaryodors:
All tests are passing.

Issue #374, opened 2023-10-10
Amino test is failing in stable branch.


primaryodors:

primaryodors:
Here are some good criteria for amino acid similarity:

```
Are both aminos' side chains hydrophilic or both hydrophobic? 50%

Both aromatic or neither aromatic? 20%
    One aromatic one conjugated non-aromatic: 10%

Both side chains same charge? 15%

Both side chains similar number of carbon atoms?
    Same number: 15%
    Differ by one: 10%
    Differ by two: 5%
```

primaryodors:
The place to implement the new criteria would be `float AminoAcid::similarity_to(const AminoAcid*)` around line 1595 of `src/classes/aminoacid.cpp`.

primaryodors:
If #373 has been merged, make sure to also uncomment lines 55-65 of `test/unit_test_express.sh` in order to reenable the amino test as part of normal unit tests.

PR #378, opened 2023-10-19
Improved amino similarity algorithm.

primaryodors:
For #374 .


primaryodors:
Motif test is failing.

primaryodors:
All current tests succeeding.

However, the propionate ligand is incorrectly protonating His180 in the OR51E2 test. It would be a good idea to expand upon the current tests for OR51E2 - propionate to ensure that I.: the His104 and His180 residues remain nonpolar and neutral, and II.: the "best-binding" algorithm finds the correct Arg262 and Gly198 contacts in the active state.

PR #382, opened 2023-10-28
Easier receptor additions.

primaryodors:
It is now easy to add proteins to the receptor repertoire, using the following steps:

- Add the protein ID and sequence in `data/sequences_aligned.txt`, using spaces to align the parts of the sequence with the neighboring proteins;
- Create an entry in `data/receptor.json`, including the `id`, `uniprot_id`, and `sequence` fields;
- Run the `www/getpdbs.php` script.

The script will automatically fill in the rest of the JSON entry and download and hydrogenate an AlphaFold model of the protein.


primaryodors:
No changes to docking code.

PR #385, opened 2023-10-31
Pull ligand toward side chains with potential for stronger binding energy.

primaryodors:
There is a constant called `lmpush` that controls motion of ligands away from atom clashes with other molecules. This PR introduces an `lmpull` constant and code to find energetically favorable interactions that could be increased by moving the ligand closer to another molecule.


primaryodors:
All tests passing.

Issue #386, opened 2023-11-05
Ligand still pulling away from binding residues. 


primaryodors:
When the primarydock app is run with the TAAR8 test config file and the `--movie` option, the ligand's N7 atom can be seen repeatedly pulling away from the Asp201 residue. This should never happen, since the intermolecular conformer search always keeps changes that make binding energy more favorable and rejects changes that result in less favorable energy, and increasing the distance between N7 and D201 results in less energetic favorability. Therefore, something is violating the principle of keeping only improvements and discarding undesired motions of molecules.

PR #387, opened 2023-11-06
Decrease amount of drift toward pocket center.

primaryodors:
This PR addresses #386 where the ligand fails to maintain contact with "best binding" residues. The effect of this code change is to increase the decay rate of the pocket center drift. In early iterations, drift is helpful because it prevents the ligand from getting tangled in clashes with side chain atoms. However, keeping the drift going strongly can prevent energetically favorable poses of the ligand in its binding site.

Additionally, drift is now suppressed if it results in a worse energy configuration for the ligand.

So far, the change seems to produce several nearly identical poses in the dock output, which is the expected behavior of molecular docking.


primaryodors:
All tests succeeding.

PR #389, opened 2023-11-11
PO numbering scheme

primaryodors:
Introduces a native PrimaryOdors numbering scheme for the olfactory receptors, based on the phylogenic tree from Zozulya et al (2001).


primaryodors:
TODO: Should have some kind of blurb on the receptors.php page that explains what all the PO numbers mean.

TBD: Should the PDB files be renamed?

TBD: Currently, each prediction generates active and inactive dock files named with the POid, and model PDBs named with the ORid. These should probably be made the same.

TODO: Check that the links all work in the context menu from the predicted column in receptor.php.

primaryodors:
Unit tests are passing, however the prediction popup contains broken links.

PR #391, opened 2023-11-15
Broken strand reconnection

primaryodors:


primaryodors:
Something is wrong with the rotations; aminos in the chain are being displaced from their neighbors with disruption of the peptide bonds.

primaryodors:
It may be worthwhile to abandon this branch and start over using the new `Molecule::evolve_structure()` function.

PR #399, opened 2023-12-01
FYG motif activation

primaryodors:
Using data from the cryo-EM models of hOR51E2, hTAAR1, and mTAAR9, it should be possible to estimate how class I and class II ORs, as well as human olfactory TAARs, might change shape upon activation.

About three quarters of all class II ORs have a Gly6.49 residue whose position is perfect for effecting the active motion of the cytoplasmic end of TMR6 by a small ψ-bond backbone rotation. Glycine is the most flexible amino acid in proteins, and the most logical site to look for a helix bend. Combining the cryo-EM motions with a Gly6.49 bend offers the current best chance of making accurate predictions about class II OR responses to ligands.


primaryodors:
All tests passing.

primaryodors:
It's doing this stupid "stack smashing" B.S. error again. No merging until that's resolved.

primaryodors:
Hoping the memory leak fixes and memory initialization fixes resolve the smashy error.

🛌🏼

primaryodors:
Would be a good idea to rerun unit_test_master before merging.

primaryodors:
All tests succeed on local development machine. However, on the server, a double free error is causing just about every test to fail. Valgrind won't even try to find where the problem is occurring. And best of all, the server is inaccessible due to an internet outage.

I'm about to go get extremely drunk. Somebody please kill me in my sleep!

PR #400, opened 2023-12-02
Cache atom geometry for performance improvement.

primaryodors:
This code change produces about sevenfold docking performance.


primaryodors:
All tests succeeded.

PR #405, opened 2023-12-14
Active Clash Compensation.

primaryodors:
When running a prediction, if any ligand-receptor clashes are observed in the active state, the directions and magnitudes (defined as ∑r(vdW) - r) of the clashes are recorded and a new temporary model PDB is generated based on adjusting the positions of the TM helices to resolve the clashes, within the constraint that the total adjustments do not exceed a fixed number of standard deviations (currently 2Σ) according to the positional variability from the cryo-EM models.


primaryodors:
All tests succeeding so far.

primaryodors:
There is a lot to this PR. However, aside from a few minor fixes, all of it is interrelated and pertinent to the application of cryo-EM models to the prediction system.

primaryodors:
All tests successful.

primaryodors:
All functionality in this PR has either been merged, is part of an open PR, or was causing unit test failures.

PR #406, opened 2023-12-16
Chime sounds at completion of predictions to audibly indicate result.

primaryodors:
Upon finishing a prediction, a sound now plays that indicates, for known empirical pairs, whether the prediction was successful (defined as correctly identifying an agonist or correctly identifying that the ligand is not an agonist) or unsuccessful, and for unknown pairs, whether the algorithm predicts an agonist or a non-agonist.

The sound is dependent on having `sox` installed. On any system without this package, the sound will not attempt to play.


primaryodors:
Should have a way to disable the sound without having to uninstall sox. Or better yet, have it be an opt-in feature that's disabled by default.

Would also like to run the sox player in the background with no output sent to the screen and no having to wait for it to finish making noise.

PR #414, opened 2023-12-22
Region clashes

primaryodors:
This PR adds information about clashes between the ligand and the TM helices to the output of `primarydock`. It is used by the FYG-active prediction method to determine whether to attempt a per-ligand customized model for the receptor's active state.


primaryodors:
Currently this is producing false positives for larger SCFA molecules (C>5) in OR51E2.

primaryodors:
All tests succeeding.

primaryodors:
The false positives for SCFAs have to do with flexibility of side chains. In the 8F76 model, the side chains of His104, Phe155, and Ile202 stay close enough together to prevent any aliphatic carbons from passing through into a large pocket between TM helices III, V, VI, and VII. But Ile202 is not constrained to the way it's positioned in 8F76, in fact the AlphaFold model has Ile202 positioned in a way that would allow aliphatic chains to reach into the large pocket. The prediction is consistent with what is known about OR51E2's structure, and the reason for the receptor to not be sensitive to SCFAs of more than 5 carbons appears to be something other than the inability of the ligands to fit inside the receptor.

primaryodors:
The dock is tearing atoms off the side chains of Gln and Glu.

primaryodors:
The stray atoms issue has not been occurring, however there is now a debug constant to help track it down if it does.

predict/methods_common.php
```diff
@@ -272,7 +274,7 @@ function prepare_outputs()
 }
 
 $multicall = 0;
-function process_dock($metrics_prefix = "", $noclobber = false)
+function process_dock($metrics_prefix = "", $noclobber = false, $no_sound_if_clashing = false)
```
primaryodors, line 277: This added feature does not seem to be working.

predict/model_accuracy.sh
```diff
@@ -1,3 +1,3 @@
 clear
-make bin/fyg_activate_or && bin/fyg_activate_or OR51E2 && bin/pepteditor predict/model_accuracy.pepd
-
+# Parameters for fyg_activate_or are copied the method_fygactive script.
```
primaryodors, line 2: Can the method_fygactive script automatically update this file whenever it is run for OR51E2?

src/classes/moiety.cpp
```diff
@@ -89,22 +97,49 @@ int Moiety::does_atom_match(Atom* a, Atom** out_matches)
...
         cursor[parens]->used = true;
 
-        for (i=1; i<n; i++)
+        for (; i<n; i++)
         {
             c = p[i];
 
             if (c <= ' ') continue;
-            if (c == '-')
+            /*if (c == '-')
```
primaryodors, line 138: Should retest with this code uncommented.

primaryodors:
The debug code is not catching the mangling problem of Gln202. Should modify the "Greek continuity" function to ensure that the moves_with array of each Bond object contains all of the residue's higher order Greek lettered atoms.

src/primarydock.cpp
```diff
@@ -2961,6 +3008,22 @@ int main(int argc, char** argv)
...
+            n = protein->get_end_resno();
+            for (i=1; i<=n; i++)
+            {
+                if (dr[drcount][nodeno].residue_clash[i])
+                {
+                    for (j=0; regions[j].start; j++)
+                    {
+                        if (i >= regions[j].start && i <= regions[j].end)
+                        {
+                            k = (i - regions[j].start) * 3 / (regions[j].end - regions[j].start);
```
primaryodors, line 3020: See if it improves accuracy to change this calculation to make the midsection smaller than 1/3 of the sequence length.

primaryodors:
OR56A5:Gln202 seems to be fixed.

predict/model_accuracy.sh
primaryodors: It would interfere with the server to have scripts always modifying tracked files. Therefore, the feature will only be enabled if the project root contains a file named `devenv`, e.g.:

`echo "yep" > devenv`

src/primarydock.cpp
primaryodors: Not really, but the change is in place.

primaryodors:
Tests are succeeding.

PR #418, opened 2023-12-29
Clash minimization work in FYG-motif activation.

primaryodors:
Hoping to see improvement in the program output...


primaryodors:
So far the optimization code is doing nothing but making the clashes worse.

PR #419, opened 2023-12-31
All atom optimization

primaryodors:
Objective: to create a function that optimizes the locations of all atoms in a molecular or multimolecular system to maximize interatomic forces and minimize clashes.


primaryodors:
Currently the new feature is failing in the primarydock app because each pose reloads the protein without deleting the old protein's atoms. Consequently, it exceeds the atom limit, but more importantly, the old atoms hang around after they're no longer valid and would interfere with the calculation.

PR #420, opened 2024-01-03
OR51 direct model

primaryodors:
Expands predict/method_directmdl to include OR51 as well as OR52.


primaryodors:
All tests passing.

primaryodors:
Prediction of isoamyl acetate fails in OR52D1 for two reasons: the directmdl method doesn't have a clash compensation section, and there is a misalignment in the EXR2 domain.

primaryodors:
All tests passing.

primaryodors:
3 fails in a row.

Redo everything from scratch.

PR #421, opened 2024-01-05
Try integrating with AutoDock.

primaryodors:


primaryodors:
Observed:
image

Predicted:
image

Done using AutoDock Vina.

primaryodors:
PrimaryDock can be made just as fast as AutoDock if the performance bottlenecks are identified and eliminated. Using our own docker means being able to adjust the finer points of things like how ligand-protein contacts are chosen etc.

Development will continue in the allow-choice-of-docker branch. This code should be combined with the current stable branch to allow the user to specify which docker to use.

PR #422, opened 2024-01-10
Allow making predictions using other molecular dockers besides PrimaryDock.

primaryodors:


primaryodors:
Pretty much every express unit test is failing. 🟥

primaryodors:
Express tests are all good. 🟩 

Three of the molecule assembly tests are failing on the server, but seem to be working fine on development machine. The server has large erroneous internal clashes for benzene and the two toluenes.

primaryodors:
[Vina ignores atom charges](https://www.researchgate.net/post/How-to-assign-correct-charge-to-Na-ions-in-a-pdbqt-file) for some preposterous reason, so metalloprotein predictions using Vina will never be as reliable as using PrimaryDock. This also means Vina cannot be used for class I ORs responding to SCFAs.

primaryodors:
All tests are passing.

PR #425, opened 2024-01-13
Staggered vs eclipsed hydrogens

primaryodors:
When bonds rotate such that atoms are placed in eclipsed positions rather than staggered, there is a loss of energetic favorability (an increase in energy) of about 11 kJ/mol. This energy cost can affect whether a receptor is sensitive to a ligand because it may be necessary for either the ligand or a side chain to flex in order to fit the ligand in the binding pocket. So even if a ligand fits the active receptor and forms good contacts, if the energy cost is greater than the difference between active and inactive binding energies, then the active conformation will not be favored and the ligand will not be an agonist.

This PR addresses the effect of eclipsing atoms and successfully distinguishes caproic and oenanthic acids as non-agonists of OR51E2 while maintaining propionic acid as an agonist.

image


primaryodors:
It is possible that the eclipsing effect may be too strong. The prediction is marking valeric acid as a non-agonist of OR51E2, when in fact it is the longest straight-chain molecule SCFA to be an agonist. It is also marking oenanthic and caprylic acids as non-agonists of OR51E1 when in fact they are agonists.

The recent addition of AutoDock Vina integration necessitated the temporary removal of the dynamic clash compensation of the active model. Bringing this feature back might resolve the aforementioned false negatives, but might also bring back false positives for OR51E2.

primaryodors:
The TAAR8 cadaverine test is failing.

primaryodors:
Tests are succeeding, but prediction now gives agonist for OR51E2 caproic acid.

PR #428, opened 2024-01-19
Enantiomers in dock and predictions.

primaryodors:
For molecules such as the gamma- and delta-lactones, ionones, alcohols, aldehydes, etc. that have multiple stereoisomers and usually exist as racemates, it is important to include all olfactory-active isomers during docking, as receptors may be more responsive to some isomers than to others. This PR establishes a means of listing out all relevant isomers of each odorant, as well as including them in a dock. The output of the `primarydock` executable identifies which isomer occurs in which pose, so that the binding energies of the isomers can be compared for that protein model.

Isomers for many compounds have also been input into `data/odorant.json` and the goal is to provide such data for every odorant or nearly every odorant that is not already listed as a single isomer, before finalizing this PR.


primaryodors:
All tests successful.

PR #431, opened 2024-01-26
Limit cfg res by type

primaryodors:
Ability to match residues by their amino acid for parameters such as STCR, FLXR, and binding pocket.


primaryodors:
All tests successful.

PR #433, opened 2024-02-03
Loose residue predictions and neighborhood grid.

primaryodors:
Combination of the `atom-neighbors` and `loose-residues` branches.


primaryodors:
Currently tracking down a segfault. Something is consistently wiping out residue 40 of OR51E2 every time running the propionate prediction.

PR #436, opened 2024-02-15
Include BW numbers in output.

primaryodors:
The output of the PrimaryDock app now gives the Ballesteros-Weinstein numbers in parentheses for all residues.


primaryodors:
Some of the tests are failing, presumably due to unexpected output formats.

primaryodors:
The tests were failing due to segfaults, possibly as a result of a linker issue. Cautiously optimistic after running `make clean` and rebuilding....

src/classes/aminoacid.h
```diff
@@ -191,6 +192,7 @@ class AminoAcid : public Molecule
     void copy_loaded_to_object(char letter, int tbdctr, AABondDef** tmpbdefs, bool proline_like);
     void find_his_flips();
 
+    Protein* protein = nullptr;
```
primaryodors, line 195: On the fence about keeping this, since we're not using it, but it's good to have that link back to the parent object.

primaryodors: Since it's here, let's go ahead and keep it, so we don't have to add it later if it becomes important.

src/classes/aminoacid.cpp
```diff
@@ -2133,6 +2134,13 @@ std::ostream& operator<<(std::ostream& os, const AminoAcid& aa)
     }
 
     os << aa.get_residue_no();
+    /*Protein* p = aa.get_parent();
```
primaryodors, line 2137: Let's remove this commented code though.

primaryodors:
Waiting on a test prediction.

PR #437, opened 2024-02-18
Manual active models

primaryodors:
A manually generated example .pepd file for creating an active state of OR5P3.

There are also fixes for the viewer's residue popup, formatting of output PDBs, and new commands for Pepteditor.

Waiting on some OR51E2 tests to find out if the code change will interfere with that receptor's successful SCFA predictions.


primaryodors:
The manual OR5P3 model has generated a successful agonist score of 6.83 for coumarin, a known strong agonist.

primaryodors:
This is now obsolete since the publication of the cryo-EM consOR models from de March et al (2024).

PR #441, opened 2024-05-30
Generational algorithm for molecular structures.

primaryodors:
Creating molecular structures now utilizes an evolutionary generational algorithm to position the atoms.

Docking now includes conformational changes of flexible rings by means of "flipping" individual atoms between two allowed locations, since anomalies in the resulting structure can be corrected using the generational algorithm.

Out of scope for this PR, but newly made possible for future development, is the ability to shape the protein around the ligand to find the mutual best fit.


primaryodors:
All tests passing. 🟩

testdata/brot.eucalyptol.approved.txt
```diff
@@ -1,3 +1,8 @@
+Bond C2-C3 can flip.
```
primaryodors, line 1: Can these really flip? Eucalyptol has a cage structure; its rings aren't free to change shape.

Issue #423, opened 2024-01-11
Bond object reciprocity.


primaryodors:
Bonds between atoms are represented in the code by a `Bond` class, which defines the arity, optimal bond length, etc.

The `Bond` object also contains two `Atom*` members, named "atom" and (foolishly) "btom". These are supposed to point to the two atoms that share the particular bond represented by the `Bond` object.

Each `Atom` object has an array of `Bond`s called `bonded_to`. The way the code works is that every bond involving any given `Atom` will appear in that `Atom`'s `bonded_to`. That means bonds are reciprocal; in an example C=O bond the C atom's `bonded_to` has a `Bond` with the "btom" pointing to the O, and the O has a `bonded_to` with the "btom" pointing to the C.

The bug is that some of the bonds are coming up as not reciprocal at runtime. The "btom" is variously either being overwritten with nonsense values or turning up null. This is causing all sorts of problems, one of which was incorrect clashes in the `mol_assem_test`s, for which the easy temporary solution was the following atrocity:

```
if (atoms[i]->is_bonded_to(atoms[j]) || atoms[j]->is_bonded_to(atoms[i]))
```

It will not be out of scope to this issue or its PR to change the name of "btom" to something less idiotic.

PR #442, opened 2024-06-01
Bond reciprocity

primaryodors:
Addresses #423 the atom pointers not always pointing to the right atoms.

Also renames the atom pointers of the Bond class to atom1, atom2.


primaryodors:
All tests succeeding. 🟩

PR #445, opened 2024-07-31
Soft Docking

primaryodors:
There's a new parameter SOFT that can be added to the config file. It will cause the various helices to pivot in order to accommodate the ligand. The pivot point is the helix's atom or residue that has the strongest binding energy with an atom outside the helix, whether this is an inter-residue disulfide bond, salt bridge, hydrogen bond, pi stack, or van der Waals bond.

Optionally, the SOFT parameter can be followed by a softness coefficient. This is a number containing a decimal point. A softness of 10 allows a maximum of 1 degree of pivot per docking iteration.

Optionally, the SOFT parameter can be followed by one or more helix numbers, e.g. 3 for TMR3, 45 for EXR2, 56 for CYT3, 7 for TMR7, etc. This number is the same as the first number of the Ballesteros-Weinstein numbering system. It is distinguished from the softness parameter by the lack of a decimal point.

When performing a SOFT dock, the output will contain an A100 measurement of the modified protein. See: http://dx.doi.org/10.1021/acs.jcim.9b00604


primaryodors:
Currently, this predicts beta-ionone as a weak agonist of OR51E2 (should be a moderate agonist). However, the most energetically favorable pose generated, with an energy of -18.0242 kJ/mol, has an A100 score of only 10.6387, indicating an inactive or partially activated receptor, albeit one that does appear able to accommodate a Golf or Gs subunit.

primaryodors:
All unit tests successful.

PR #447, opened 2024-08-02
Automated Homology Modeling

primaryodors:
Automates the homology modeling process by dynamically creating and executing the necessary scripts. Requires MODELLER to generate the models.

Usage:
```
php -f hm/dohm.php RCPID
```

...replacing RCPID with the receptor ID, for example OR1A1.

A few sample HM-generated active-state PDBs are included, however they are not yet producing good predictions.


primaryodors:
All tests passing.

primaryodors:
Tests succeeding.

The predictions made using the new HM structures are coming up consistently wrong, except for the OR51E2 and TAAR8 tests. Also the ligand in the OR51E2 test is oriented differently than in the cryo-EM model. Nevertheless, the scope of this PR is automating generation of HM structures, not producing correct predictions. Any further development would only constitute feature creep.

PR #448, opened 2024-08-05
Taar8 bbalign fix

primaryodors:
Fixes a problem with the "best binding" algorithm that was causing the HM active model of TAAR8 to fail to form good contacts with cadaverine.


primaryodors:
All tests succeeding.

PR #454, opened 2024-08-19
Dock optimizations

primaryodors:
Various sundry optimizations for docking and predictions:

- Improvements in CS search;
- Improvements in BB search;
- Fix hydrogenation of histidines;
- Soft docking applied only to active-state phase of prediction;
- Crash fixes;
- HM templates now include bovine rhodopsin;
- Improvements for metal coordination;
- Reduced hydrogen bonding of acceptors in aromatic rings.

Assay feature: There is now a script called `predict/assay.php` that, when given an odorant name as a command line argument, will go through all the GPCRs running predictions for that odorant. After several days, longer if homology models must be built, the `predict/dock_results.json` file will contain a full set of predictions for that ligand, and there will be dock files and models in `output/`.

New search mode: CP. This mode takes a PDB file containing a ligand, and applies the 3D atom coordinates from that model to the ligand before running dock iterations. Both the source PDB and the dock PDB should first be aligned to each other, e.g. with the `UPRIGHT` and `BWCENTER` commands in pepteditor.


primaryodors:
All tests succeeding.

primaryodors:
All tests successful.

primaryodors:
82% accuracy in predicting a set of known receptors for leaf alcohol. (Green and red dots manually added to screenshot to show correct vs. incorrect predictions.)

image

primaryodors:
Active dock from successful OR2M3 prediction:

image

PR #455, opened 2024-09-09
Equilibrium formula

primaryodors:
Receptor-ligand activity shold not be determined by a straightforward comparison of active vs. inactive energies, but rather by an equilibrium formula that takes into account the equilibrium proportions of active vs. inactive receptor states based on the difference between ligand binding energies.

https://www.masterorganicchemistry.com/2023/08/02/equilibrium-constant-delta-g-calculations-organic-chemistry/


primaryodors:
All tests are passing.

primaryodors:
Only small changes to docking code; tested OR51E2 and TAAR8 without issue. Running express unit tests.

PR #460, opened 2024-09-16
Multiple pocket centers.

primaryodors:
Config files can now specify multiple `CEN RES` parameters, and PrimaryDock will try all of them during docking. This is useful for receptors that have different binding pockets depending on the ligand, as orthosteric and allosteric binding sites can be defined and attempted in predictions.

The `data/binding_pocket.json` file can specify multiple pockets using an array, as it now does for OR51*.

Also includes an improvement to the metal coordination code.

All tests are passing.


primaryodors:
All tests are passing. Also, the OR51E2 propionate test is pretty consistently placing the carboxylate group in a vertical position now, with good bonds to both Arg262 and Ser258, like it is in the 8F76 model:

image

PR #466, opened 2024-09-22
Std::vectors have got to go.

primaryodors:
It is not conducive to development when sections of code, that have worked fine all along, randomly start throwing untraceable segfaults deep inside std::vector code.

All std::vector objects are to be replaced with C++ arrays and any attempt to use the std::vector class should throw a compiler error.


primaryodors:
Docks are noticeably slower for some reason.

All tests failing.

primaryodors:
PR rejected.

PR #468, opened 2024-09-25
WCLASH new dock result metric.

primaryodors:
Adds a new field to the JSON output of predictions. WCLASH represents the smallest value of any pose's worst atom-atom clash. Each pose's "worst atom clash" value is already used to filter dock results, however this new "best worst atom clash" metric can be used in prediction scoring algorithms.

Additionally, the atom clash limit has been increased, there is a fix for loading PDBs that have duplicate atom records, docks no longer cram nonpolar hydrogens into the centers of histidine rings, and the Predictions page of the web app has a more simplified look.


primaryodors:
Good results on OR51E2 predictions with butyric, valeric, isovaleric, and caproic acids.

primaryodors:
All tests passing.

PR #471, opened 2024-10-11
Fix prediction failures caused by incorrect predock ligand position.

primaryodors:
Corrects the position of the ligand prior to dock iterations. This will fix prediction failures such as OR51E1 - pelargonic acid which have been failing under the new cavity based docking due to placing the carboxyl group away from Arg264(6.59) and the end of the hydrophobic tail deep inside the receptor instead of near Ala206(5.46).


primaryodors:
Reverted the SEARCH parameter back to CS by default, BB for TAARs. CF was giving too many failed predictions from bad active poses.

PR #473, opened 2024-10-21
Bridging fix

primaryodors:
Fixes BRIDGE and STCR directives in config files so that they take effect as expected.

Note that these directives are run once per pose, so different output poses might show bridged side chains in slightly different positions in 3D space.


primaryodors:
All tests passing.

PR #474, opened 2024-10-21
Cavity-matching search.

primaryodors:
Conformer search that works by matching one ligand atom to one cavity partial, then performing 3DOF rotations to maximize containment.


primaryodors:
All tests passing. 🟩

primaryodors:
Ligand internal energy from atom eclipses shouldn't preclude binding to the protein. Here's are examples of experimental structures where the ligand is in a highly bent or crumpled state: https://www.rcsb.org/3d-view/1ACA/0 https://www.rcsb.org/3d-view/1B56/1 https://www.rcsb.org/3d-view/3WEH/1

primaryodors:
🟩

PR #477, opened 2024-11-05
Homology models from new consensus ORs.

primaryodors:
A [new study](https://doi.org/10.1038/s41586-024-08126-0) has released long-awaited cryo-EM models of Class II ORs bound to agonists and G proteins. The models are of consensus ORs of the OR1, OR2, OR4, and OR51 families. This PR introduces active-state models based on the new consensus ORs.


primaryodors:
There's a code change in here somewhere that broke the OR51E2 and TAAR prediction tests.

primaryodors:
🟩

PR #478, opened 2024-11-08
Fix errors in alignments for homology models.

primaryodors:
The OR1A subfamily has a deletion between the EXR2 HXXCD/E motif and the TMR5 SY motif, and the OR10H subfamily has an insertion in the same place. This PR fixes the alignments used for homology modeling.

Also included is the capacity to correctly predict hazelnut pyrazine as an agonist for OR5K1. T279(7.42) forms an hbond with atom N7 of the pyrazine ring, and the aliphatic substitutions make contacts with L104(3.32) and L255(6.51).


primaryodors:
🟩

PR #482, opened 2024-11-17
Merge overlapping cavities.

primaryodors:
Formerly, the cavity_search utility sometimes generated separate cavities with considerable overlap that would be better conceptualized as one larger cavity. This PR fixes that problem by looking for overlapping generated cavities and merging them together.


primaryodors:
🟩

PR #483, opened 2024-11-21
Output dock energy as difference between protein-bound ligand and water-solvated ligand energies.

primaryodors:
Based on [Floriano et al (2000)](http://dx.doi.org/10.1073/pnas.97.20.10712) in which the "binding energies were calculated as the difference between the energy of the ligand in protein and in solution," this PR replaces the previous behavior of modeling only protein-ligand energies with the new behavior of first computing the ligand's potential interaction energy with liquid water, then subtracting that value from the dock result energy.


primaryodors:
🟩

PR #485, opened 2024-12-09
Active-only predictions

primaryodors:
Empirical receptor-ligand activity does not negatively correlate to inactive-state dock energy nor to the difference between active and inactive state dock energies. It is apparently not necessary to perform the inactive dock in order to make predictions.

This PR removes the inactive dock from predictions and adds two new result metrics: ligand surface polarity (LSP) and ligand surface receptor binding (LSRB). In theory, LSP represents how much of the ligand's surface is available to hydrogen bonding in extracellular epithelial fluid, and if LSRB exceeds LSP, then the ligand is likely to prefer the receptor's binding pocket rather than an aqueous solution. The new prediction algorithm uses LSP and LSRB in the formula that produces the dock score.

There are also fixes to vdW interactions, isomers of citronellic acid, cavity search, odorant structures, and the cavity_fit utility has been rewritten.


primaryodors:
OR51E2 propionate test failed by < 0.02 kJ/mol. All other tests are 🟩.

primaryodors:
🟩

primaryodors:
🟩

primaryodors:
OR51E2:
image

PR #487, opened 2024-12-13
LSRB math fix

primaryodors:
Fixes a math error in the LSRB calculation.

Also includes minor adjustments to the interaction code and docking algorithms, and adds a unit test for geraniol in OR1A1.


primaryodors:
🟩

PR #488, opened 2024-12-14
Visualize LSRB

primaryodors:
Adds visualization of the ligand surface's proximity to the receptor's side chains in order to see where in space the contributions to the LSRB value are coming from.


primaryodors:
🟩 

There is a known bug in this version: when viewing a dock file in 3D, after turning on the LSRB visualization, if the user then clicks to open an SDF, the viewer will fail to display the SDF leaving only a blank view and a Please Wait message. The error happens somewhere in the third party NGL code. This does not happen if LSRB visualization is turned off when opening the SDF.

primaryodors:
Bug fixed.

PR #489, opened 2024-12-15
LSRB part II

primaryodors:
Fixes the LSRB calculation to be spatially accurate.

Out of scope for this PR would be some way to spatially estimate how stable the ligand is to "slipping out", i.e. are there any holes in the LSRB big enough for the ligand to slide through and break contact with the receptor. In a future PR, this could be calculated and called something like "stability".


primaryodors:
🟩

PR #493, opened 2024-12-22
Subtract binding for certain residues from dock score.

primaryodors:
Many false positive ligands, i.e. those predicted to be agonists yet empirically known as non-agonists, form strong bonds with side chains involved in stabilizing the receptor's active state. This can be seen in a dock of (Z)-3-hexen-1-ol in OR2B11, empirically a non-agonist pair, in which the ligand binds to E184(45.51), likely disrupting its hydrogen bond to Y263(6.55). Since this bonding between side chains is believed to stabilize the active state, an odorant that disrupts the bond would not be an agonist.

This PR introduces code that subtracts binding energy to Y6.55 and D/E45.51 if present. Other important side chains can be added as they are found by continuing efforts of docking ligands and examining false positives.


primaryodors:
🟩

PR #494, opened 2024-12-22
Reuse best pose from past pose attempts.

primaryodors:
This PR introduces a small change that instructs Primarydock to save the ligand and side chain positions from the most successful pose attempt and reuse these positions in subsequent pose attempts. Whereas before this change, the OR1A1 geraniol test was generating 1 to 2 output poses, it now generates 7-8 output poses, nearly all of which make a strong contact between the ligand and the important Asn109(3.37) residue.

Even if one generated pose only makes a weak contact to an important side chain, subsequent poses based on the weak pose are likely to develop stronger contacts.

The downside to doing this is that it masks the possibility of finding multiple dissimilar positions for a given ligand in a given receptor. To mitigate this disadvantage, a `reuse_pose_probability` value is defined in constants.h, that can be changed to any value between 0 and 1 inclusive. Increasing this value will produce more output poses that resemble each other, while decreasing it will produce fewer successful poses with more variation between them.


primaryodors:
🟩

PR #495, opened 2024-12-23
Enhancement to prevent atom clashes.

primaryodors:
Interaction class now requires worst atom clash to be below the global threshold before allowing improvements in the attractive force. What this means is during docking, when testing a new positional or rotational conformer candidate, the algorithm will not allow changes that increase any atom clashes beyond the global atom clash limit, even if the change would increase the total energy favorability.


primaryodors:
🟩

PR #496, opened 2024-12-23
Separate best-pose caches for isomers of ligand.

primaryodors:
The recent PR #494 failed to take into account that if a ligand has multiple isomers defined as `ISO` params in the config file, as is the case with chiral molecules docked as a racemate, then the isomer chosen by the docker for any given pose attempt is not necessarily the same isomer that was saved as the "best pose".

This PR fixes the issue by assigning each isomer its own "best pose" up to a theoretical maximum of 64 isomers. It also increases the minimum number of pose attempts in the prediction scripts for molecules that are docked with isomers so that each isomer has a fair chance at generating successful output poses. Finally, it adds a unit test for OR2M3 and its strongest agonist 3-mercapto-2-methylpentan-1-ol.


primaryodors:
🟩

PR #497, opened 2024-12-23
Mirrored docking of achiral ligands.

primaryodors:
Certain odorants, such as citral, cinnamaldehyde, and benzophenone, ideally would have planar structures however the models in the SDF files are bent in order to accommodate steric effects. As such, these molecules in the real world at ambient temperature can freely flip back and forth between [two mirror image conformations](https://www.nature.com/articles/s41598-021-99606-0) and can therefore fit into a binding pocket even if it can only accommodate one of the two conformers. But the docking code did not have a facility for docking both conformers until now.

PrimaryDock will now test each ligand to determine whether it is chiral and, if not, randomly mirror image the ligand for roughly half of all docking pose attempts. This ensures that no matter which conformer is a better fit for the binding pocket, it will be included in the dock.


primaryodors:
🟩

PR #498, opened 2024-12-24
Store disruptible residues in an array.

primaryodors:
The list of internal protein contacts that can be disrupted by ligand binding, is now an array stored in `methods_common.php`. It no longer requires customization in `method_directmdl.php` but can still be overridden there. Residues can be added and removed in one place and the scripts will take care of synchronizing the list everywhere it has an effect.

Active models for OR5K have been updated to include a hydrogen bond between Y6.55 and D/E45.51. Doubtless there are other models that also should be similarly updated.


primaryodors:
🟩

primaryodors:
🟩

PR #499, opened 2024-12-25
LSRB expansion

primaryodors:
The goal of this PR is to increase the LSRB for ligands that are well surrounded by receptor side chains. For example, an odorant might fit snugly into the binding pocket of one receptor while a different receptor leaves one side of it wide open, yet it can have close to the same LSRB in both receptors. Ideally, in the former case, the LSRB should be close to 1 and in the latter case more like 0.6.


primaryodors:
🟩 

But I'd still like to think of a way to fill in LSRB areas that are "flanked" by high value vertices while leaving large empty spaces empty.

primaryodors:
🟩

primaryodors:
🟩

PR #500, opened 2024-12-28
Favor poses where ligand is less bent.

primaryodors:
Adds internal clashes and eclipses to the dock iterations so that the ligand naturally tends to straighten out if it can. The result is more output poses with better energy and stronger contacts between interacting moieties.

Also includes enhanced docking constants and miscellaneous fixes.


primaryodors:
🟩

PR #501, opened 2024-12-31
Fix internal contacts in OR models.

primaryodors:
This PR ensures that the active models for class II ORs have the important Y6.55~D/E45.51 hydrogen bond, as well as the disulfide bond between TMR3 and the EXR2 loop, and also the proximity of the hydroxyl groups of Y5.58 and Y7.53.

Obviously not all class II receptors have all of these residues, so each structural feature will only be present in the proteins that have the requisite residues.

Out of scope for this PR would be to ensure that the aromatic ring of Y6.55 is positioned more or less vertically with the terminal hydrogen pointing up at one of the oxygen atoms of 45.51, however I am suffering from an advanced case of theresonlyoneofmeism and the CS method doesn't properly handle catechols docked in OR10G* models.


primaryodors:
🟩

PR #502, opened 2025-01-01
Preserve stay-close atoms of ligand's best pose.

primaryodors:
The recently added feature to reuse the running best pose candidate #494 did not preserve the "stay-close" atoms of the ligand and protein. This means if a ligand atom in the best pose is strongly contacting one side chain of the receptor, a subsequent pose attempt could disrupt this by assigning the ligand atom to a different side chain, reducing the pose attempt's effectiveness. This PR fixes that.

Also included: if reusing a previous pose, it is not necessary to re-run the pre-positioning code to generate a pre-position that will only be discarded.


primaryodors:
🟩

PR #503, opened 2025-01-01
Prevent ligand outside protein.

primaryodors:
This update does 2 things:

1.) Prevent reusing poses where the ligand is not inside one of the known ligand binding pockets;
2.) Prevent output of poses where the ligand is not inside one of the known ligand binding pockets.


primaryodors:
🟩

PR #506, opened 2025-01-13
Side chain hydrogenation

primaryodors:
Fixes a bug in the hydrogenation of side chain atoms.


primaryodors:
This is failing unit tests all over the place.

PR #509, opened 2025-01-16
Dock auditing.

primaryodors:
Adds the capability to audit the dock iterations by generating an output file listing the changes made to the ligand and protein in each iteration.

Also fixes an intermittent problem with the TAAR8 test by making the Best Binding algorithm the default for docking. Of all the algorithms, Best Binding has a track record of producing the greatest number of higher quality output poses.

Also included are updated active receptor models made using MODELLER's DOPEHRLoopModel with restraints for maintaining the Y6.55-D/E45.51 and Y5.58-Y7.53 atom proximities, and the new corrected hydrogenation code. Regrettably, DOPEHRLoopModel does not seem to preserve cystine cross links, so a restraint has been added and the thiol hydrogens are removed by the dynamic .hm.pepd script.


primaryodors:
One of the new active models is wrong and it's causing unit test fails. OR51E2 has a His180 side chain that should be placed "up" away from Ser258, towards the extracellular end of TMR5, to form a sort of "roof" to the ligand binding site. Instead, the new active model has His180's side chain pointing in the direction of Ser258, blocking the ligand's access to its polar binding residues Ser258 and Arg262.

In and of itself this would be an easy fix - perform a bond rotation on the CA-CB of His180 and save the updated PDB. However, this is actually indicative of a bigger problem. Firstly, how did OR51E2 become changed into an incorrect structure? It should be a copy of the 8F76 cryo-EM structure from RCSB. Nothing in the PDB file indicates that this structure was generated using HM. Secondly, irrespective of how the OR51E2 model was produced, could whatever changed this structure have also changed other receptor structures making them incorrect as well?

primaryodors:
The change happens in the `BWCOPY` command in Pepteditor. This is incorrect behavior because `BWCOPY` should never change any part of the 3D structure, only the labeling and numbering of regions.

primaryodors:
Problem turned out to be in the `HYDRO` command.

All 409 OR and TAAR active models must now be redone.

primaryodors:
🟩

src/primarydock.cpp
```diff
@@ -3263,7 +3288,11 @@ int main(int argc, char** argv)
             if (pdpst == pst_best_binding) ligand->movability = (MovabilityType)(MOV_CAN_AXIAL | MOV_CAN_RECEN | MOV_CAN_FLEX);
 
             freeze_bridged_residues();
+            if (auditfn.length()) audit = fopen(auditfn.c_str(), (pose==1 && !nodeno) ? "wb" : "ab");
```
primaryodors, line 3291: Should check to make sure the file opened successfully and warn the user (through cerr) if not.

PR #512, opened 2025-01-22
Prevent sidechain clashes

primaryodors:
Adds a new parameter for the sidechain clashes delta to measure decreases or increases in clashes between protein atoms resulting from docking.

Ideally, side chain clashes should never increase and there is code in this PR to prevent increases, but they are still coming from somewhere. Measures have been put into place to prevent SC clash increases in:
- the dock algorithm itself, i.e. `Molecule::conform_molecules`;
- `Molecule::conform_atom_to_location`, used by the `ATOMTO` param of the config file;
- `Molecule::minimize_internal_clashes`.

Finding the cause of remaining clash increases and solving that problem will help to reduce the number of false positives, since most false positives currently involve a clash increase (but so do some correct predictions).


primaryodors:
Filtering out all clash increases from dock results breaks the OR1A1 geraniol test.

image

Unfortunately, the right-for-the-wrong-reason result must remain in place for now, since the `stable` branch cannot be allowed to start consistently failing.

primaryodors:
🟨 

Master unit tests succeeded, but not consistently. Running concurrently on two machines, the CS test for OR51E1-caprylate succeeded only once, and the OR1A1 geraniol prediction succeeded only once.

CS mode is likely to be removed in the near future. OR1A1 - geraniol seems to be less reliable now that fewer side chain clashes are getting through to the dock results.

PR #514, opened 2025-02-05
mclashables in bond rotations

primaryodors:
Limits flexions of protein side chains so that clashes are not allowed to increase between residues.


primaryodors:
🟩

PR #516, opened 2025-04-13
Enhanced Soft Docking

primaryodors:
Soft docking no longer relies on static pivot points, but instead attempts to maintain internal contacts while moving around soft regions.


primaryodors:
🟩

primaryodors:
🟥 

Beta ionone fails to pair its ketone oxygen with OR51E2:Arg262 in BB test.

Something is really seriously slowing down the dock to an unacceptable degree, and gprof is giving confusing output.

I can't give this project the time it deserves. It was necessary to drop all but one of my classes and there is no hope of continuing to move forward in my degree program. Applying to hundreds of jobs has yet to produce so much as an interview. Playing fake taxicab all day is mathematically incapable of bringing in enough income to meet basic expenses.

Someone else can fork this project and see if they can beat it into submission. All I ask is to send me a list of all the receptors and receptor combinations you can correlate to a perceptual note. Assuming I still have electricity by then.

PR #517, opened 2025-06-04
Soft docking fixes

primaryodors:
Soft docking now makes use of internal contacts within the protein, i.e. residues whose side chains form hydrogen bonds, salt bridges, etc. These internal contacts are used partially to restrain the ways the parts of the protein are allowed to move around, and partly to estimate the energy cost of expanding to accommodate a large ligand.

Cystine cross links are not included in the measured internal contacts, therefore for GPCRs the EXR2 loop is not allowed to move relative to TMR3.

Some bugs in the soft docking code are also resolved by this PR.

Additionally this includes a rewrite of the Best Binding algorithm. The new methodology is based on finding compatible contacts between the protein and ligand that are matched for relative distances.


primaryodors:
🟩

PR #519, opened 2025-06-07
Bug fixes for new BB algorithm.

primaryodors:
Allows ligands with only 2 pairable groups. Also suppresses the BB algorithm if a ligand has only one group, e.g. hydrogen sulfide. Also includes conjugated rings as pairable groups, and limits the size of paired groups so that e.g. cinnamaldehyde, whose heavy atoms are all conjugated together, isn´t represented as one big whole-molecule group.

Also fixes a problem with SDF generation of enantiomers, and suppresses the attempted use of cavity files if they haven´t been generated.


primaryodors:
🟩

PR #520, opened 2025-06-09
Remove obsolete algorithms

primaryodors:
Also includes miscellaneous small fixes... and a not so small fix, an SDF for an enantiomer of onion pentanol that obabel is currently failing to generate.


primaryodors:
🟩

PR #522, opened 2025-06-10
Fix segfault on single ligand atom.

primaryodors:
Also adds a unit test of prediction of H<sub>2</sub>S in OR2T11.


primaryodors:
🟩

PR #527, opened 2025-06-13
Prevent skipping iterations before last reappearance.

primaryodors:
As a performance feature, when the dock detects the ligand pose has not changed in several iterations and is energetically favorable, it will skip to the end of iterations for that pose. Obviously, this is not desired behavior if DISAPPEAR and REAPPEAR parameters have been set, since it is important to continue the dock until all residues have been un-hidden.

This PR introduces a check to make sure the dock completes a minimum number of iterations each pose, that number being equal to the highest numbered iteration from all DISAPPEAR and REAPPEAR params.


primaryodors:
🟩

primaryodors:
🟩

PR #529, opened 2025-06-18
Ligand to explore open spaces.

primaryodors:
This image shows heptanal, a non-agonist, in OR3A1:

image

The hydrophobic tail is sticking into a large cavity where it has lots of space to flex and is not held snugly by nearby residues like we typically see with agonists. 

This PR introduces code to look for such loose hydrophobic parts of ligands and, rather than gluing them to the walls of large cavities, causes them to float far from side chains so that the overall dock energy is less favorable. In real-world molecules, these floating groups would decrease the stability of a ligand-bound protein and make it less likely that the ligand is an agonist.

The consequence of this PR is that it correctly predicts heptanal as a non-agonist for OR3A1, when before it was predicting an agonist.


primaryodors:
🟥

PR #533, opened 2025-06-24
Prediction test panel

primaryodors:
Ultimately, the predictions in `unit_test_master` will be superseded by a panel of mandatory predictions that must succeed in every PR alongside the unit tests, with new predictions becoming mandatory as development increases prediction accuracy.

```
php -f predict/panel.php mandatory
```

For now, the panel contains a selection of receptor+agonist pairs that represent a cross section of olfactory receptors and aroma classes, and would make for tremendously valuable cryo-EM models in case anyone involved in such research happens to be reading this...

This PR also implements a long overdue fix for ionic bond energies.


primaryodors:
🟥 

The only test failing is the OR2M3 test, and it is also failing in the `use-correct-sign-internally` branch, which seems to be otherwise succeeding. Even "successful" OR2M3 tests have been generating only single pose results with clashes between the ligand and the metal coordination residues. The OR2M3 model is probably incorrect, and the test should not be a requirement for merging.

primaryodors:
🟩

PR #538, opened 2025-07-19
Quit button

primaryodors:
- Pressing Q during docking now stops the dock and outputs whatever poses it has found so far;
- When the docker finds a suitable output pose, the progress bar turns green;
- Web app: the receptor page now has a "juxtapose" mode showing active and inactive models with a different color for each OR family;
- Web app: there is a new `seqs.php` page, still in development, that shows sequence alignments;
- The .ali files for HM now align the n.50 residues of each helix at the 100th character;
- A few new nodes in the btree;
- Dynamic soft-contact elasticity;
- Optimize internal contacts in each final result pose;
- If BB fails to assign pairs, it retries ignoring compatibility between ligand atoms and side chains;
- New parameters in `/bin/ic` app;
- New `/bin/qc` app for finding quaternary-structure contacts, e.g. between receptor and Gα subunit;
- Bond rotation capability in SMILES viewer.


primaryodors:
🟥 

The only fail is the TAAR8-cadaverine test. The ligand did not make strong contacts with Asp111, so the solvated algorithm deemed it less strongly bound than if it were in water solution. As the solvated algorithm is slated to be replaced soon, this is not a show stopper. Re-running the test to see if it happens again and if it succeeds, this PR can be code reviewed and merged.

primaryodors:
🟩

PR #540, opened 2025-07-27
Capability to perform full protein scans for ligand binding sites.

primaryodors:
The `test/bb_test` application now offers the capacity to scan an entire protein for candidate binding sites for a given ligand. Running it with the `full` argument causes it to ignore any `cen` residues and instead attempt "best-binding" searches on every spatially plausible combination of compatible residues for binding the ligand's functional groups. The results are filtered to exclude intervening residues that would clash with the ligand, however this filter isn't flawless.

The following example illustrates the syntax of the new feature:

`test/bb_test full pdbs/OR51/OR51E2.active.pdb sdf/beta-ionone.sdf`

The full scan has identified a possible allosteric binding site in OR51E receptors for non-acidic ligands. When run for OR51E2 and its agonist beta-ionone, as shown above, it identified a pocket near the residues with Ballesteros-Weinstein numbers 3.20, 3.22, 3.25, 3.26, 45.39, 45.43, 45.45, and 45.50, between extracellular loops 1 and 2, of the right size and shape to fit not only ionones but also certain OR51E1 agonists such as menthol, methyl salicylate, and methyl eugenol that fit poorly or not at all in the orthosteric binding pocket between helices III, V, and VI.

The new tentative binding site was added to `data/binding_pocket.json`. Since soft docking normally requires all non-helical regions to be excluded from the calculation, the prediction scripts were modified to retain the involved residues, whose regions aren't defined as "soft" and therefore will remain stationary relative to each other. Interestingly, two of the residues of the hypothesized allosteric pocket, 3.25 and 45.50, form the highly conserved disulfide bond found in nearly all GPCRs. Since ORs have low tertiary structure stability and therefore suffer from low cell surface expression, and since both [agonist](https://doi.org/10.1016/j.bbrc.2025.151459) and [antagonist](https://doi.org/10.3390/ijms26041458) ligands have been shown to increase OR cell surface expression likely due to chaperone effects on OR structure, there may be a connection between the allosteric pocket and receptor structural stabilization.

Running a prediction of β-ionone in OR51E2, using the new hypothesized pocket, failed to produce a dock result, however all three of the aforementioned OR51E1 agonists successfully docked in the EXR1/EXR2 pocket of that receptor.

The OR2M3 test has been removed from the master unit tests due to repeated failures. The copper binding site and Thr105 residue are too close together in the active model, causing a poor fit for 3-mercapto-2-methylpentan-1-ol.


primaryodors:
The current version of the full scan doesn't identify the TMR3-EXR2 pocket for beta ionone because the earlier result depended on Ile91 whose CB atom points away from the pocket center, causing that residue to be excluded from consideration. The current version of the code takes a very long time to scan OR51E2 for ionone.

primaryodors:
The current version finds Phe168 and Glu169 as binding residues of adenosine in the active ADORA2A model, but fails to identify Ser277 and His278.

primaryodors:
🟩

PR #542, opened 2025-08-02
Adjustments to total dock energy

primaryodors:
Soft contact anomaly is now included in total docking energy.

Inactive dock is once again included in predictions.

Additionally, there is now a computed charge repulsion effect when a negatively charged (Asp or Glu) side chain falls near a thiolated Cys participating in metal binding. The ligand binding pockets of OR2T receptors change very little between active and inactive states, so the ligand energy is similar for both, resulting in failed predictions. 

All OR2M/T/V receptors conserve D4.56 near the copper binding site and a sort of a nook between G3.37, G4.53, and (in all but two cases) G4.57, where the carboxylate group of D4.56 can tuck into. The "nook" in the inactive models is smaller and slightly distorted, making the "tuck" less likely without clashing, while in the active HM models based on cryo-EM results of consensus ORs, the "nook" is cavernous enough to accommodate the "tuck", reducing the proximity, and therefore the charge repulsion, between D4.56 and the thiolated C5.42. Even the carvone-bound consOR2 [model](https://www.rcsb.org/structure/8UY0) itself features this nook, with D154(4.56) tucked into it.

Therefore, when a metal ion is present, in the active state, PrimaryDock now estimates this charge repulsion and adds it to the total dock energy of the output pose, resulting in the ability to distinguish agonist thiols from inactive alcohols in OR2T11:

image


primaryodors:
🟩

primaryodors:
🟩

Issue #543, opened 2025-08-02
Dynamics estimates.


primaryodors:
The OR51E1 caprylate test fails unless the R6.59 residue is deleted from inactive docking. Reason being the caprylate molecule can span from the EXR2 end of the binding pocket of the inactive receptor, out through empty space, to the isolated R6.59 side chain, with a more favorable dock energy than sitting in its designated place in the active model. In reality, the former arrangement is much less dynamically stable than the latter, as both the ligand and R6.59 side chain can easily jostle loose, and caprylate is an agonist, not an inverse agonist as predicted.

Doing nanosecond by nanosecond dynamics simulations is beyond the scope of PrimaryDock; the calculations involved would take weeks. Instead, there are ways to estimate dynamic stability of the receptor+ligand complex. One of these is already partially available: the ligand surface receptor binding (LSRB) metric, a number between 0 and 1 that's already computed in every dock, is a measure of how well the ligand is "gripped" by the receptor's binding pocket. Experimental X-ray and cryo-EM structures of an agonist in a GPCR usually show the ligand nestled tightly into the pocket, so in theory a ligand with a high LSRB in the active state model should be more likely to be an agonist. However, some agonists have low LSRB values and some non-agonists have a higher LSRB and a more favorable dock energy than agonists of the same receptor.

If the LSRB concept be applied to individual side chains, one would find that OR51E1:R6.59 has a moderately high LSRB in the active state and an almost zero LSRB in the inactive state. That, coupled with a less dramatic difference in the ligand LSRB, would favor the active state once the two LSRB values are multiplied together and used as a multiplier for that residue's ligand binding energy. 

An even better calculation could also take into account the residue's flexion probability (FP), which for arginine is 0.254. The likelihood that R6.59 would be stable would be decreased by its high FP, but that effect would be modulated in the active state by its LSRB.

Differences in FP could also explain why OR1A2 is not a broadly tuned receptor like its close relative OR1A1. The side chain of OR1A1:Asn109(3.37) is thought to be an important polar moiety for ligand binding, and asparagine's FP is a moderate 0.105, giving it an intermediate level of dynamic stability as it juts unsupported into the binding pocket with a low LSRB. Most of the dynamic stability would be in the hydrophobic ligand-receptor interactions, which is why OR1A1 is sensitive to alkenes (e.g. limonene) that lack polar groups, and not to small molecular weight alcohols that cannot fill the binding pocket in a dynamically stable manner.

[...]

PR #544, opened 2025-08-07
Dynamic stability estimations

primaryodors:
Implements some of the proposed changes in #543 , q.v. for background. Namely, each ligand binding residue's stability estimate acts as a multiplier for its ligand binding energy, and the total energy is output both with and without this correction applied. Predictions therefore base their decisions on stability-mitigated dock energies.

Prediction uses the delta algorithm again, under the theory that ligands favoring active or inactive receptor states will tend to be agonists or non-agonists, respectively.

A new metric, the RTSC or receptor tertiary stabilization coefficient, is introduced. It is based on pairs of non-contiguous regions such as TMR3-TMR5 and EXR2-TMR6. For each pair, the ligand's binding energies to each region are used for calculating the ligand's ability to "glue" those regions together. The sum of all these values represents the ligand's predicted ability to stabilize the receptor in its modeled state (active or inactive), and the RTSC values for the two states are used by the prediction algorithm.

Also fixes a few of the Python scripts and adds new scripts for `data/gensdf` and `predict/emp`.

The receptor phylogenic tree has been cleaned up, placing a few "straggler" receptors with their families, and most notably uniting OR10 within the OR6/10/11 clade.

Prediction assays can now be done with a btree node as a second argument. To find the btree for a clade of receptors, the user can navigate the tree view of the receptors page and then copy the `bt` param from the URL.

Fixed a bug in Best Binding where the third "rotisserie" step formerly used contact residues to define the axis, resulting in unrecoverable clashes for long-chain SCFAs (e.g. C8) in Class I ORs. Now the algorithm uses ligand molecular features to define the axis of rotation, and multiple successful poses are output for caprylate in OR51E1.

Includes a workaround for a bug in OR2T receptors where a thiol ligand oftentimes finds very successful poses during dock iterations, but then loses energy favorability later. When a suitable strongly bound pose is found, iterations finish early. This results in some very fast (~2-3 min) docks of small molecule thiols in OR2T*.

Brought back anisotropy of metals. Quality of metal-coordinated thiolate docks has improved.

Predictions list now shows dock score.


primaryodors:
The AlphaFold model of OR2AT4 is misaligned in TMR6, resulting in a lack of the conserved Y6.55~D45.51 hydrogen bond. The active model from HM has this hbond, however the active dock of sandalore disrupts it. For some reason, the dock fails to identify that as a soft contact anomaly.

